PR #7, opened 2023-02-28
vch: enable C++ inclusion support

sa-kib:
Specify namespace so that vch library can be used from C++ code.

### DIFF
--- a/vch/include/vch.h
+++ b/vch/include/vch.h
@@ -10,6 +10,10 @@
 #include <zephyr/xen/public/xen.h>
 #include <zephyr/xen/public/io/libxenvchan.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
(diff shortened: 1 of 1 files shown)


oleksiimoisieiev:
LGTM

vch/include/vch.h
```diff
@@ -10,6 +10,10 @@
 #include <zephyr/xen/public/xen.h>
 #include <zephyr/xen/public/io/libxenvchan.h>
 
+#ifdef __cplusplus
```
al1img, line 13: Many c++ applications use c libraries. But not all c libraries use `extern "C"` directive. So, IMHO it should be resolved on the application side.

sa-kib: most of zephyr modules use it like this, so it looks like expected thing to be done by C library

firscity:

PR #8, opened 2023-02-28
xenstore: add xss_rm() routine

sa-kib:
This shall allow to do XenStore clean-ups on dom0 side, e.g. vchannel server must remove its XS entries when closing connection.


oleksiimoisieiev:
Looks good, But from my standpoint commit message should be more detailed. For example I don't understand why do you need this function to be exported and why previous features didn't require it.

sa-kib:
> Looks good, But from my standpoint commit message should be more detailed. For example I don't understand why do you need this function to be exported and why previous features didn't require it.

commit description is updated

firscity (approved):

PR #9, opened 2023-03-08
Fixups fixing build errors and warnings in Zephyr.

oleksiimoisieiev:
This MR includes 2 fixups:
1) Fixing build error if libfdt is not enabled
2) Fixing warning of returning value from void function


firscity (approved):
Acked-by: Dmytro Firsov <dmytro_firsov@epam.com>

PR #11, opened 2023-03-09
xen-dom-mgmt: Removing pages from Dom0 physmap before mapping

oleksiimoisieiev:
Allocated pages should be released before mapping Domain pages because it should be replaced in P2M without releasing any RAM that may be associated with page_addr.


xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -269,6 +269,34 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
+	for (i = 0; i < nr_pages; i++) {
+		rc = xendom_remove_from_physmap(DOMID_SELF, mapped_pfns[i]);
+		if (rc < 0)
+			break;
+	}
+
+	return rc;
+}
+
+static int add_pages_to_physmap(int domid,
```
firscity, line 285: This function doing something different to how it called.

firscity (commented):
My point was to add 2 helper functions, that can be easily modified to use with extended regions:
1. Let it be called `xen_pfn_t *xen_get_mapping_pfns(uint64_t nr_pages)`. This function should (currently) allocate region of `nr_pages` with `k_aligned_alloc()`, remove them from physmap and return as xen_pfn_t array to a caller. Caller then may use it for mapping DomU pages for DomU kernel/dtb copying using standard `xendom_add_to_physmap_batch()` call.
2. `int xen_put_mapping_pfns(xen_pfn_t *pfns, uint64_t nr_pages)` - this function should remove pages from Dom0 physmap and fill hole in address space with `xendom_populate_physmap()` to make it returnable to Zephyr allocator. Also it may return error code if something gone wrong.

In future it will help to migrate to extended regions without changing anything in domain management code.

oleksiimoisieiev:
> My point was to add 2 helper functions, that can be easily modified to use with extended regions:
> 
>     1. Let it be called `xen_pfn_t *xen_get_mapping_pfns(uint64_t nr_pages)`. This function should (currently) allocate region of `nr_pages` with `k_aligned_alloc()`, remove them from physmap and return as xen_pfn_t array to a caller. Caller then may use it for mapping DomU pages for DomU kernel/dtb copying using standard `xendom_add_to_physmap_batch()` call.
> 
>     2. `int xen_put_mapping_pfns(xen_pfn_t *pfns, uint64_t nr_pages)` - this function should remove pages from Dom0 physmap and fill hole in address space with `xendom_populate_physmap()` to make it returnable to Zephyr allocator. Also it may return error code if something gone wrong.
> 
> 
> In future it will help to migrate to extended regions without changing anything in domain management code.

Agreed. Those functions can be used in allocate_magic_pages as well.

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,107 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
-void load_dtb(int domid, uint64_t dtb_addr, const char *dtb_start, const char *dtb_end)
+static int xen_rm_pages_from_physmap(uint64_t nr_pages, void *mapped_addr)
 {
-	int i, rc;
-	void *mapped_dtb;
-	uint64_t mapped_dtb_pfn, dtb_pfn = XEN_PHYS_PFN(dtb_addr);
-	uint64_t dtb_size = dtb_end - dtb_start;
-	uint64_t nr_pages = ceiling_fraction(dtb_size, XEN_PAGE_SIZE);
+	int i, rc = 0;
+	xen_pfn_t mapped_pfns[nr_pages];
```
xakep-amatop, line 278: (note) from Zephyr Code Guidelines:
```
Rule 18.8  Required  Variable-length array types shall not be used
```

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,107 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
+		rc = xendom_remove_from_physmap(DOMID_SELF, mapped_pfns[i]);
+		if (rc < 0)
+			break;
+	}
+
+	/*
+	 * After this Dom0 will have memory hole in mapped_domu address,
+	 * needed to populate memory on this address before freeing.
+	 */
+	rc = xendom_populate_physmap(DOMID_SELF, 0, nr_pages, 0, mapped_pfns);
```
xakep-amatop, line 297: (question major) should you pass `i` or `i+1` instead of `nr_pages` here?
the above `xendom_remove_from_physmap` call can fail and we exit from the loop. I don't know anything about `xendom_populate_physmap` function and its logic, but can be a major issue here...

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,107 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
+	if (rc < 0)
```
xakep-amatop, line 298: (minor/major) pls, add brackets here

I prefer to use curly braces even for one line, problems begin when:

* somebody comments or deletes the line under if statement
* somebody adds a few new lines and forget to add curly braces
* somebody adds multiline macro and uses it and if statement without curly braces

----
```
Add braces to every if, else, do, while, for and switch body, even for single-line code blocks. Use the --ignore BRACES flag to make checkpatch stop complaining.
```
https://docs.zephyrproject.org/3.2.0/contribute/guidelines.html#coding-style

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,107 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
-	uint64_t mapped_dtb_pfn, dtb_pfn = XEN_PHYS_PFN(dtb_addr);
-	uint64_t dtb_size = dtb_end - dtb_start;
-	uint64_t nr_pages = ceiling_fraction(dtb_size, XEN_PAGE_SIZE);
+	int i, rc = 0;
+	xen_pfn_t mapped_pfns[nr_pages];
+	xen_pfn_t mapped_addr_pfn;
+
+	mapped_addr_pfn = xen_virt_to_gfn((uint64_t)mapped_addr);
+
+	for (i = 0; i < nr_pages; i++)
```
xakep-amatop, line 283: (cosmetic) comparison between signed and unsigned types, `int < uint64_t`

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,107 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
+		mapped_pfns[i] = mapped_addr_pfn + i;
+
+	/* Needed to remove mapped Domaim pages from Dom0 physmap */
+	for (i = 0; i < nr_pages; i++) {
+		rc = xendom_remove_from_physmap(DOMID_SELF, mapped_pfns[i]);
+		if (rc < 0)
+			break;
```
xakep-amatop, line 290: (question) Should we print warn about error here? it would be good to know how often it can happen, should we add some stats here?

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,107 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
 
-	mapped_dtb = k_aligned_alloc(XEN_PAGE_SIZE, XEN_PAGE_SIZE * nr_pages);
-	if (!mapped_dtb)
-		return;
+	addr = k_aligned_alloc(XEN_PAGE_SIZE, XEN_PAGE_SIZE * nr_pages);
+	if (!addr)
+		return -ENOMEM;
 
-	mapped_dtb_pfn = xen_virt_to_gfn((uint64_t)mapped_dtb);
+	LOG_INF("Allocated %llu pages (%llu), mapped_domd = %p", nr_pages,
```
xakep-amatop, line 330: (question) could you please add `domid ` too?

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,107 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
+			XEN_PAGE_SIZE * nr_pages, addr);
+	memset(addr, 0, XEN_PAGE_SIZE * nr_pages);
+
+	rc = xen_rm_pages_from_physmap(nr_pages, addr);
+	if (rc < 0)
+		return rc;
```
xakep-amatop, line 336: (major) looks like you forget to call `k_free` for `addr` here and below in case of errors

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,107 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
+	rc = xendom_add_to_physmap_batch(DOMID_SELF, domid,
+			  XENMAPSPACE_gmfn_foreign, nr_pages,
+			  indexes, mapped_pfns, err_codes);
+	if (rc < 0)
+		goto err;
+
+	*mapped_addr = addr;
+	return rc;
+err:
+	ret = xen_put_mapping_pfns(nr_pages, addr);
```
xakep-amatop, line 354: (major) it is very bad idea to override `rc` after error, lets imagine the next situation:
1. you are inside `load_dtb`
2. `mapped_dtb`variable is  uninitialized and has some stack garbage
3.  you call `xen_get_mapping_pfns` and pass to function pointer of `mapped_dtb`
4.  you get error during `xendom_add_to_physmap_batch` call
5. you override `rc` here by `xen_put_mapping_pfns`
6. returned from function with `0`
7. use `mapped_dtb` with some garbage inside

there are a few solutions here:
* initialize `mapped_dtb` variable and and check it after returning from `xen_get_mapping_pfns` function (you can even initialize it inside of `xen_get_mapping_pfns` function, but in case of error)
* don't override `rc` inside `xen_get_mapping_pfns` in case of first error

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,107 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
+	if (ret)
+		LOG_ERR("Unable to free recourses rc = %d", ret);
+	return rc;
```
xakep-amatop, line 357: (major) looks like we need to call here `k_free` in order to avoid memory leak

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -313,41 +386,24 @@ void load_dtb(int domid, uint64_t dtb_addr, const char *dtb_start, const char *d
 	rc = xen_domctl_cacheflush(0, &cacheflush);
 	LOG_DBG("Return code for xen_domctl_cacheflush = %d", rc);
```
xakep-amatop, line 387: (question) should you do something else in case of error?
I think in case of error better to use LOG_ERR or LOG_WRN

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,107 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
+			  indexes, mapped_pfns, err_codes);
+	if (rc < 0)
+		goto err;
+
+	*mapped_addr = addr;
+	return rc;
+err:
+	ret = xen_put_mapping_pfns(nr_pages, addr);
+	if (ret)
+		LOG_ERR("Unable to free recourses rc = %d", ret);
```
xakep-amatop, line 356: (note) you already have error logs inside `xen_put_mapping_pfns` function, I think this is a little bit redundant

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -313,41 +386,24 @@ void load_dtb(int domid, uint64_t dtb_addr, const char *dtb_start, const char *d
...
-	/* Needed to remove mapped DomU pages from Dom0 physmap */
-	for (i = 0; i < nr_pages; i++) {
-		rc = xendom_remove_from_physmap(DOMID_SELF, mapped_pfns[i]);
-		if (rc < 0) {
-			LOG_ERR("Error while removing physmap (rc=%d)", rc);
-			goto out;
-		}
+	rc = xen_put_mapping_pfns(nr_pages, mapped_dtb);
+	if (rc < 0) {
+		LOG_ERR("%d: xen_put_mapping_pfns rc = %d\n", __LINE__, rc);
```
xakep-amatop, line 391: (note) you already have error logs inside `xen_put_mapping_pfns` function, I think this is a little bit redundant

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -313,41 +386,24 @@ void load_dtb(int domid, uint64_t dtb_addr, const char *dtb_start, const char *d
...
+		return;
```
xakep-amatop, line 392: (cosmetic) this line is redundant

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -410,27 +452,18 @@ int probe_zimage(int domid, uint64_t base_addr, uint64_t image_load_offset,
 	rc = xen_domctl_cacheflush(0, &cacheflush);
 	LOG_DBG("Return code for xen_domctl_cacheflush = %d", rc);
 
-	/* Needed to remove mapped DomU pages from Dom0 physmap */
-	for (i = 0; i < nr_pages; i++) {
-		rc = xendom_remove_from_physmap(DOMID_SELF, mapped_pfns[i]);
-		if (rc < 0)
-			goto out;
+out:
```
xakep-amatop, line 455: (cosmetic) unused label

xen-dom-mgmt/src/xen-dom-mgmt.c
xakep-amatop, line 336: (major) looks like you forgot to call `k_free` here

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: Oh. Thanks for the point.

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: xendom_remove_from_physmap can fail if pages was already removed. Then we need to populate memory to Dom0 to fill up the hole even if not all pages were present in the physmap. That's my understanding. Lets summon @firscity to be sure.

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: +

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: +

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: +

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,107 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
 	int err_codes[nr_pages];
-	struct xen_domctl_cacheflush cacheflush;
+	void *addr;
 
-	mapped_dtb = k_aligned_alloc(XEN_PAGE_SIZE, XEN_PAGE_SIZE * nr_pages);
-	if (!mapped_dtb)
-		return;
+	addr = k_aligned_alloc(XEN_PAGE_SIZE, XEN_PAGE_SIZE * nr_pages);
+	if (!addr)
+		return -ENOMEM;
```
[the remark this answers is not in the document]
oleksiimoisieiev: I think we don't. Error code is enough. Higher level will print error and error code

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: +

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: Hm. And what should be done in this case from your standpoint?

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: +

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: +

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: +

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: no. I don't need k_free here for the memory that was just allocated.

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,120 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
+	return rc;
+}
+
+static int xen_put_mapping_pfns(int nr_pages, void *mapped_addr)
+{
+	int rc;
+
+	rc = xen_rm_pages_from_physmap(nr_pages, mapped_addr);
+
+	k_free(mapped_addr);
```
LKomaryanskiy, line 318: I have a question: should we do k_free here even if xen_rm_pages_from_physmap exits with an error code?

xen-dom-mgmt/src/xen-dom-mgmt.c
xakep-amatop: Could you please explain it? I haven't seen where this memory is used in case of error and where it is freed

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: removed

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: We can't free addr. because it's goes outside on line 361. Also we use in to calculate mapped_addr_pfn. Why should I free it?

oleksiimoisieiev: Oh. now I got your point. Sorry , my bad. I'll fix it

xen-dom-mgmt/src/xen-dom-mgmt.c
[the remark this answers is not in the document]
xakep-amatop: actually, we don't print error messages in caller function only dbg logs and I'm sure that we won't enable dbg log level by default

xen-dom-mgmt/src/xen-dom-mgmt.c
firscity: Population of physmap pages should not be called in function `xen_rm_pages_from_physmap()` at all. Part after line 300 should be removed. You are removing pages mapping and then populate them back with RAM, when you need to just remove mapping.

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: Yes we do. We have to release allocated memory regardless of whether xen_rm_pages call was failed or not

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,120 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
+	}
+
+	return rc;
+}
+
+static int xen_put_mapping_pfns(int nr_pages, void *mapped_addr)
+{
+	int rc;
+
+	rc = xen_rm_pages_from_physmap(nr_pages, mapped_addr);
```
firscity, line 316: Here should be 2 separate calls remove page mapping and populate physmap. If both are OK, you can return these addresses to memory allocator- with `k_free()`. Not checking rc code may break Zephyr allocator.

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,120 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
+		return -ENOMEM;
+
+	LOG_INF("DOM(%d): Allocated %llu pages (%llu), mapped_addr = %p", domid,
+			nr_pages, XEN_PAGE_SIZE * nr_pages, addr);
+	memset(addr, 0, XEN_PAGE_SIZE * nr_pages);
+
+	rc = xen_rm_pages_from_physmap(nr_pages, addr);
+	if (rc < 0) {
+		goto err;
+	}
```
firscity, line 345: It should be the end of this function. All other action should be performed by caller outside. This function should only return region suitable for foreign memory mapping.

firscity (changes requested):

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -33,7 +33,7 @@
 #include <xenstore_srv.h>
 
 LOG_MODULE_REGISTER(xen_dom_mgmt);
-
+#define DOMAIN_PAGES_MAX 256
```
firscity, line 36: Also, why did you chose this value?

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,120 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
 	return modbase;
 };
 
-void load_dtb(int domid, uint64_t dtb_addr, const char *dtb_start, const char *dtb_end)
+static int xen_rm_pages_from_physmap(uint64_t nr_pages, void *mapped_addr)
```
xakep-amatop, line 275: (note) if we have so small max amount of pages `DOMAIN_PAGES_MAX 256`, we can avoid using so big types for `nr_pages` here and below

xen-dom-mgmt/src/xen-dom-mgmt.c
firscity: > Yes we do. We have to release allocated memory regardless of whether xen_rm_pages call was failed or not

No, we are not. If we return broken region to Zephyr allocator, we will receive exception when someone will try to write something there after `k_malloc/k_aligned_alloc`. We should keep this region as broken on our side, or try to fix it and then return to allocator.

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -272,34 +272,120 @@ static uint64_t get_dtb_addr(uint64_t rambase, uint64_t ramsize,
...
 {
-	int i, rc;
+	int rc = 0;
+	unsigned int i;
+	xen_pfn_t mapped_pfns[DOMAIN_PAGES_MAX];
```
xakep-amatop, line 279: (note) I hope that this approach in a future doesn't overflow stack frame
as for me we can omit using of arrays at all, we need only 3 variables:
* start address
* number of items
* size of one item

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: Suggest me better one.

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: I think this can be done on additional MR

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: @firscity why did you made it uint64_t originally?

xen-dom-mgmt/src/xen-dom-mgmt.c
LKomaryanskiy: @xakep-amatop currently, I'm working on proper return code handling and error messages. So, I think, @oleksiimoisieiev can leave code that is not related to his changes and I will fix this in other places, like this.

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: I'd rather leave arrays so I'm able to call add_to_physmap_batch to operate with all buffer and don't split it to many add_to_physmap calls.

xakep-amatop: I think it is too risky to grab so much memory on stack frame and we need to redo all stuff related to arrays

xakep-amatop: maybe some global variables with locks
or some arrays in struct context with locks

oleksiimoisieiev: I think this can be done on additional MR.

xen-dom-mgmt/src/xen-dom-mgmt.c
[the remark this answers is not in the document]
oleksiimoisieiev: @LKomaryanskiy  works on this right now as he mentioned below.

xen-dom-mgmt/src/xen-dom-mgmt.c
lorc: You can add some comment that justify this value. "Number 256 was chosen randomly" is fine :)

xen-dom-mgmt/src/xen-dom-mgmt.c
lorc: @xakep-amatop, FYI: `add_to_physmap_batch` is much more efficient than mapping page-by-page.

xen-dom-mgmt/src/xen-dom-mgmt.c
lorc: I am curious how next part will work if there are some pages left in physmap.... You need to investigate this.

xen-dom-mgmt/src/xen-dom-mgmt.c
xakep-amatop: it depends on how it realized under hood
but if you say that batch function is efficient, I believe you

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: > "Number 256 was chosen randomly" is fine :)
You made my day. )))) We've discussed with Dima this part and decided that this part should be refactored copletely.

LKomaryanskiy:
@oleksiimoisieiev thank you for your PR. PR with similar functionality  has been merged:
https://github.com/xen-troops/zephyr-xenlib/pull/19
So you can close this PR.

PR #14, opened 2023-03-16
xen-dom-mgmt: print rambase & kernbase in hex

sa-kib:
Fix inconvenient addresses output, preserve decimals for sizes only.


firscity (approved):

PR #15, opened 2023-03-16
xen-dom-mgmt: place image at text_offset

sa-kib:
According to arm64 boot protocol [1] zimage can specify offset from RAM base at which it shall be placed. However current loader implementation is placing image at rambase only -- by mapping base_addr to mapped_domd and copying image there we got start of image placed at rambase. Yet after that we return correct start address of rambase + text_offset.

So far it worked because of zero text_offset of domd images, so it wasn't really accounted for. But as soon as image with non-zero text_offset provided -- the execution of domain fails, basause we jump to address beyond start of image.

This is fixed by calculating load address, with text_offset taken into account, and mapping this load address to local buffer. Also a better naming is provided.

[1] https://docs.kernel.org/arm64/booting.html?highlight=text_offset#call-the-kernel-image


lorc:
Looks good.

firscity (approved):
Thank you.

PR #16, opened 2023-03-20
Xrun implementation to zephyr-xenlib.

oleksiimoisieiev:
Introducing xrun api to the zephyr xenlib. Xrun library allows application to provide bundle with runtime spec and start provided services as the Domains.


xen-shell-cmd/CMakeLists.txt
```diff
@@ -3,6 +3,6 @@
 add_library(XEN_SHELL INTERFACE)
 
 zephyr_library()
-zephyr_library_sources(src/xen_shell.c src/xen_cmds.c)
+zephyr_library_sources(src/xen_shell.c src/xen_cmds.c src/xrun_cmds.c)
```
xakep-amatop, line 6: (question) is it possible to build xen-shell without enabled `CONFIG_XRUN`?
if yes, should you add some checks here? I mean to add this file as a library source only in case when `CONFIG_XRUN` is enabled

xrun/include/storage.h
```diff
@@ -0,0 +1,22 @@
...
+#ifndef XENLIB_XRUN_STORAGE_H
+#define XENLIB_XRUN_STORAGE_H
+
+#include <sys/types.h>
+#include <zephyr/types.h>
+
+#ifdef DEBUG
+int lsdir(const char *path);
+int write_file(const char *path, const char *name, char *buf,
+					 size_t size);
```
xakep-amatop, line 15: (cosmetic) it isn't necessary to move this to another line, Zephyr allows maximum 100 symbols in one line

xrun/include/storage.h
```diff
@@ -0,0 +1,22 @@
...
+#endif
+
+ssize_t read_file(const char *path, const char *name,
+				  char *buf, size_t size, int skip);
```
xakep-amatop, line 19: (cosmetic) the same as above

xen-shell-cmd/src/xrun_cmds.c
```diff
@@ -0,0 +1,149 @@
+// SPDX-License-Identifier: Apache-2.0
```
xakep-amatop, line 1: (minor) pls, avoid using of one line comment style

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -402,11 +443,17 @@ int probe_zimage(int domid, uint64_t base_addr, uint64_t image_load_offset,
 	if (rc < 0)
 		goto out;
 
-	LOG_DBG("Zephyr DomD start addr = %p, end addr = %p size = 0x%llx",
-	       img_start, img_end, domd_size);
+	LOG_DBG("Zephyr Domain start addr = %p, binary size = 0x%llx\n",
```
xakep-amatop, line 446: (cosmetic) new line isn't needed, Logging subsystems add it automatically

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -373,7 +414,7 @@ int probe_zimage(int domid, uint64_t base_addr, uint64_t image_load_offset,
 	}
 
 	dtb_addr = get_dtb_addr(base_addr, KB(domcfg->mem_kb), load_addr,
-				 domcfg->img_end - domcfg->img_start, fdt_size);
+				 domd_size, fdt_size);
 	if (!dtb_addr)
```
xakep-amatop, line 418: (major) looks like here we can have memory leak and few lines we below we also have return without memory free
```
if (mapped_domd == NULL)
		return 0;
```

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -343,25 +343,66 @@ int probe_zimage(int domid, uint64_t base_addr, uint64_t image_load_offset,
...
+	rc = domcfg->get_image_size(domcfg->image_info, &domd_size);
+	if (rc < 0 || domd_size == 0) {
+		LOG_ERR("Error calling get_image_size rc: %d\n", rc);
+		return rc;
+	}
+
+	domd_size -= image_load_offset;
+	nr_pages = ceiling_fraction(domd_size, XEN_PAGE_SIZE);
+
+	mapped_pfns = k_calloc(nr_pages, sizeof(xen_pfn_t));
```
xakep-amatop, line 384: (question) why can't you allocate one contiguous chunk of memory here instead of three calls? 
in this case:
* you can avoid calling so often alloc function
* you can avoid calling so often free function
* you will have less code inside this function

Are you afraid that we don't have one big non fragmented chunk of memory?

xrun/include/xrun.h
```diff
@@ -0,0 +1,15 @@
+/* SPDX-License-Identifier: Apache-2.0
+ *
+ * Copyright (c) 2023 EPAM Systems
+ */
+
+#ifndef XRUN_H_
+#define XRUN_H_
+
+int xrun_run(const char *bundle, int console_socket, const char *container_id);
```
Eabramov84, line 9: Due to it is an interface please add comments and return value descriptions.

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -343,25 +343,66 @@ int probe_zimage(int domid, uint64_t base_addr, uint64_t image_load_offset,
...
 
-	struct zimage64_hdr *zhdr = (struct zimage64_hdr *)img_start;
-	uint64_t load_addr = base_addr + zhdr->text_offset;
-	uint64_t load_pfn = XEN_PHYS_PFN(load_addr);
+	struct zimage64_hdr zhdr;
+	uint64_t load_addr;
+	uint64_t load_pfn;
+
+	if (!domcfg->load_image_bytes || !domcfg->get_image_size) {
+		LOG_ERR("Image callback is not set\n");
```
xakep-amatop, line 361: (typo) callback**s**

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -402,11 +443,17 @@ int probe_zimage(int domid, uint64_t base_addr, uint64_t image_load_offset,
...
-	       img_start, img_end, domd_size);
+	LOG_DBG("Zephyr Domain start addr = %p, binary size = 0x%llx\n",
+	       mapped_domd, domd_size);
 
 	/* Copy binary to domain pages and clear cache */
-	memcpy(mapped_domd, img_start, domd_size);
+	rc = domcfg->load_image_bytes(mapped_domd, domd_size,
+					 image_load_offset, domcfg->image_info);
+	if (rc < 0) {
+		LOG_ERR("Error calling load_image_bytes rc: %d\n", rc);
```
xakep-amatop, line 453: (cosmetic) new line isn't needed

xen-shell-cmd/src/xrun_cmds.c
```diff
@@ -0,0 +1,149 @@
...
+#include <zephyr/shell/shell.h>
+
+const char *get_param(size_t argc, char **argv, char opt)
+{
+	int pos;
+
+	for (pos = 1; pos < argc; pos++)
+		if (argv[pos][0] == '-' && argv[pos][1] == opt) {
+			/* Take next value after option */
+			pos++;
```
xakep-amatop, line 17: (question minor/major) is it possible that user after option enter a few spaces or tabs? should you handle it here?

xen-shell-cmd/src/xrun_cmds.c
```diff
@@ -0,0 +1,149 @@
...
+	container_id = get_param(argc, argv, 'c');
+	bundle = get_param(argc, argv, 'b');
+	socket = atoi(get_param(argc, argv, 's'));
+
+	if (!container_id) {
+		shell_error(shell, "Invalid containerid passed to run cmd\n");
+		return -EINVAL;
+	}
+
+	return xrun_run(bundle, socket, container_id);
```
xakep-amatop, line 44: (question major) is it ok to have bundle equal to NULL?

xen-shell-cmd/src/xrun_cmds.c
```diff
@@ -0,0 +1,149 @@
...
+	const char *bundle;
+	int socket = 0;
+
+	if (argc < 5 || argc > 7) {
+		return -EINVAL;
+	}
+
+	container_id = get_param(argc, argv, 'c');
+	bundle = get_param(argc, argv, 'b');
+	socket = atoi(get_param(argc, argv, 's'));
```
xakep-amatop, line 37: (minor) I believe that you should perform checking of ret value from `atoi, an error can happen inside

xen-shell-cmd/src/xrun_cmds.c
```diff
@@ -0,0 +1,149 @@
...
+	return NULL;
+}
+
+static int xrun_shell_run(const struct shell *shell, size_t argc, char **argv)
+{
+	const char *container_id;
+	const char *bundle;
+	int socket = 0;
+
+	if (argc < 5 || argc > 7) {
```
xakep-amatop, line 31: (question minor) why do you need this checks at all here and below functions?
you already provide an amount of mandatory options to cmd constructor:
```
/**
 * @brief Initializes a shell command with arguments.
 *
 * @note If a command will be called with wrong number of arguments shell will
 * print an error message and command handler will not be called.
 *
 * @param[in] syntax	 Command syntax (for example: history).
 * @param[in] subcmd	 Pointer to a subcommands array.
 * @param[in] help	 Pointer to a command help string.
 * @param[in] handler	 Pointer to a function handler.
 * @param[in] mand	 Number of mandatory arguments including command name.
 * @param[in] opt	 Number of optional arguments.
 */
#define SHELL_CMD_ARG(syntax, subcmd, help, handler, mand, opt) \
```

xen-shell-cmd/src/xrun_cmds.c
oleksiimoisieiev: This is the requirement of checkpatch. It requires different format of SPDX for *.c and *.h files.

xen-shell-cmd/src/xrun_cmds.c
oleksiimoisieiev: This implementation doesn't care about command line format. So if user wants some special chars in command line - they should provide the following format: param="test" and make sure it will be accepted by json.

xrun/include/storage.h
oleksiimoisieiev: checkpatch in zephyr allows only 80 and I've checked all of my patches. If it doesn't fit format - I would have got error.

xen-shell-cmd/src/xrun_cmds.c
```diff
@@ -0,0 +1,149 @@
...
+		" Resume container\n"
+		" Usage: resume -c <container_id>\n",
+		xrun_shell_resume, 3, 0),
+	SHELL_CMD_ARG(state, NULL,
+		" Show container state\n"
+		" Usage: state -c <container_id>\n",
+		xrun_shell_state, 3, 0),
+	SHELL_SUBCMD_SET_END);
+
+SHELL_CMD_ARG_REGISTER(xrun, &subcmd_xrun, "XRun commands", NULL, 2, 0);
```
xakep-amatop, line 149: (note) looks like we have to set mandatory number of args to 3 instead of 2
it should be min of mandatory from `subcmd_xrun` array

xen-shell-cmd/src/xrun_cmds.c
```diff
@@ -0,0 +1,149 @@
...
+
+	if (argc < 3 || argc > 4) {
+		return -EINVAL;
+	}
+
+	container_id = get_param(argc, argv, 'c');
+
+	if (!container_id) {
+		shell_error(shell,
+					"Invalid containerid passed to resume cmd\n");
```
xakep-amatop, line 97: (cosmetic) something strange with formatting

xrun/src/storage.c
```diff
@@ -0,0 +1,158 @@
+// SPDX-License-Identifier: Apache-2.0
```
xakep-amatop, line 1: (cosmetic) pls, avoid using of one line comments here and in other files

xen-shell-cmd/src/xrun_cmds.c
xakep-amatop: it is strange because zephyr sources don't have such strange combination
are you tacking about checkpatch for which project Linux or Zephyr?

xrun/src/storage.c
```diff
@@ -0,0 +1,158 @@
...
+
+ssize_t read_file(const char *path, const char *name,
+				  char *buf, size_t size, int skip)
+{
+	struct fs_file_t file;
+	ssize_t rc;
+	int ret;
+	char fname[MAX_PATH_LEN];
+
+	snprintf(fname, MAX_PATH_LEN, "%s/%s", path, name);
```
xakep-amatop, line 26: (note) this function can return negative error value

xrun/src/storage.c
```diff
@@ -0,0 +1,158 @@
...
+					 char *buf, size_t size)
+{
+	struct fs_file_t file;
+	int rc, ret;
+	char fname[MAX_PATH_LEN];
+
+	snprintf(fname, MAX_PATH_LEN, "%s/%s", path, name);
+
+	fs_file_t_init(&file);
+	rc = fs_open(&file, fname, FS_O_CREATE | FS_O_RDWR);
```
xakep-amatop, line 135: (question) why isn't enough to use `FS_O_WRITE`?

xrun/include/storage.h
xakep-amatop: maybe we use different checkpatch scripts, because for me 100 symbols work good

xrun/src/xrun.c
```diff
@@ -0,0 +1,404 @@
...
+		return ret;
+
+	return unregister_container_id(container_id);
+}
+
+int xrun_state(const char *container_id)
+{
+	struct container *container = get_container(container_id);
+
+	if (!container)
```
xakep-amatop, line 400: (cosmetic) pls. use curly braces even for one line under loop/if statements it is a requirement from Zephyr Code Guidelines here and above

xrun/src/xrun.c
```diff
@@ -0,0 +1,404 @@
...
+	struct container *container = register_container_id(container_id);
+
+	if (!container)
+		return -EINVAL;
+
+	bytes_read = read_file(bundle, "config.json", config,
+					XRUN_JSON_SIZE_MAX, 0);
+	if (bytes_read < 0) {
+		LOG_ERR("Can't read config.json ret = %ld\n", bytes_read);
+		return bytes_read;
```
xakep-amatop, line 315: (question) should you unregister container in case of any error or only after domain_create?

xrun/src/xrun.c
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	parse_config_json(config, bytes_read, &container->spec);
```
xakep-amatop, line 318: (minor) unhandled status

xen-shell-cmd/CMakeLists.txt
oleksiimoisieiev: +

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: +

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: fixed. thanks

xen-dom-mgmt/src/xen-dom-mgmt.c
oleksiimoisieiev: +

xrun/src/storage.c
oleksiimoisieiev: +

xrun/src/storage.c
oleksiimoisieiev: What if file doesn't exists? this was taken from sample.

xrun/src/xrun.c
```diff
@@ -0,0 +1,428 @@
...
+err:
+	k_spin_unlock(&container_lock, key);
+
+	return container;
+}
+
+static int unregister_container_id(const char *container_id)
+{
+	struct container *container = get_container(container_id);
+	k_spinlock_key_t key;
```
firscity, line 178: I'm not sure if it is OK, to declare variables after function calls.

xrun/src/xrun.c
```diff
@@ -0,0 +1,428 @@
...
+	return 0;
+}
+
+static int load_image_bytes(uint8_t *buf, size_t bufsize,
+			uint64_t image_load_offset, void *image_info)
+{
+	ssize_t res;
+	struct container *container;
+
+	if (!image_info) {
```
firscity, line 197: Should we verify buf too?

xrun/src/xrun.c
```diff
@@ -0,0 +1,428 @@
...
+
+	return (size == 0) ? -EINVAL : 0;
+}
+
+static int fill_domcfg(struct container *container)
+{
+	ssize_t res;
+	struct xen_domain_cfg *domcfg = &container->domcfg;
+
+	domcfg->mem_kb = 4096;
```
firscity, line 234: Should all of these values be hardcoded here? Looks pretty strange

xrun/src/xrun.c
```diff
@@ -0,0 +1,428 @@
...
+	domcfg->max_evtchns = 10;
+	domcfg->max_vcpus = 1;
+	domcfg->gnt_frames = 32;
+	domcfg->max_maptrack_frames = 1;
+
+	domcfg->nr_iomems = 0;
+
+	/* irqs = domd_irqs, */
+	domcfg->nr_irqs = 0;
+#if defined(CONFIG_XENVM_USE_GIC_V3)
```
firscity, line 245: This config is related to Dom0 Zephyr, not to created domain. AFAIK, if HW runs on GICv3, created domain may use both GICv2/GICv3.

xrun/src/xrun.c
```diff
@@ -0,0 +1,428 @@
...
+}
+
+int xrun_run(const char *bundle, int console_socket, const char *container_id)
+{
+	int ret = 0;
+	ssize_t bytes_read;
+	char *config;
+	struct container *container = register_container_id(container_id);
+
+	if (!container) {
```
firscity, line 311: Should we verify bundle too?

xrun/src/xrun.c
```diff
@@ -0,0 +1,428 @@
...
+		ret = -ENOMEM;
+		goto err;
+	}
+
+	bytes_read = read_file(bundle, "config.json", config,
+					CONFIG_XRUN_JSON_SIZE_MAX, 0);
+	if (bytes_read < 0) {
+		LOG_ERR("Can't read config.json ret = %ld\n", bytes_read);
+		ret = bytes_read;
+		k_free(config);
```
firscity, line 326: it may be another label before `err` and be reused in next condition too

xen-shell-cmd/src/xrun_cmds.c
```diff
@@ -0,0 +1,137 @@
...
+		return -EINVAL;
+	}
+
+	rc = xrun_state(container_id, &state);
+	if (!container_id) {
+		shell_error(shell, "Unable to get specs for the container\n");
+		return -EINVAL;
+	}
+
+	printf("%s state is %d\n", container_id, state);
```
firscity, line 109: I'm not sure about `printf()`, it probably should be `shell_print()` here.

xen-shell-cmd/src/xrun_cmds.c
oleksiimoisieiev: finally fixed

xen-shell-cmd/src/xrun_cmds.c
```diff
@@ -0,0 +1,137 @@
...
+
+	container_id = get_param(argc, argv, 'c');
+
+	if (!container_id) {
+		shell_error(shell, "Invalid containerid passed to state cmd\n");
+		return -EINVAL;
+	}
+
+	rc = xrun_state(container_id, &state);
+	if (!container_id) {
```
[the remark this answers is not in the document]
oleksiimoisieiev: fixed

xen-shell-cmd/src/xrun_cmds.c
oleksiimoisieiev: fixed

xrun/include/storage.h
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+#ifndef XENLIB_XRUN_STORAGE_H
+#define XENLIB_XRUN_STORAGE_H
+
+#include <sys/types.h>
+#include <zephyr/types.h>
+
+#ifdef DEBUG
+int lsdir(const char *path);
```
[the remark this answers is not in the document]
oleksiimoisieiev: fixed

xrun/src/storage.c
```diff
@@ -0,0 +1,170 @@
...
+#include <stdio.h>
+
+#include <zephyr/device.h>
+#include <zephyr/fs/fs.h>
+#include <zephyr/fs/littlefs.h>
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+
+#include <storage.h>
+#define MAX_PATH_LEN 255
```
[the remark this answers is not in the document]
oleksiimoisieiev: Yes. it was taken from lfs.h. Dropped this define and used LFS_NAME_MAX

xrun/src/storage.c
```diff
@@ -0,0 +1,170 @@
...
+
+ssize_t read_file(const char *path, const char *name,
+				  char *buf, size_t size, int skip)
+{
+	struct fs_file_t file;
+	ssize_t rc;
+	int ret;
+	char fname[MAX_PATH_LEN];
+
+	ret = snprintf(fname, MAX_PATH_LEN, "%s/%s", path, name);
```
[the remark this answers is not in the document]
oleksiimoisieiev: fixed

xrun/src/storage.c
```diff
@@ -0,0 +1,170 @@
...
+	return rc;
+}
+
+ssize_t get_file_size(const char *path, const char *name)
+{
+	int rc;
+	struct fs_dirent dirent;
+	char fname[MAX_PATH_LEN];
+
+	rc = snprintf(fname, MAX_PATH_LEN, "%s/%s", path, name);
```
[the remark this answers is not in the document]
oleksiimoisieiev: +

xrun/src/storage.c
```diff
@@ -0,0 +1,170 @@
...
+}
+
+int write_file(const char *path, const char *name,
+					 char *buf, size_t size)
+{
+	struct fs_file_t file;
+	int rc, ret;
+	char fname[MAX_PATH_LEN];
+
+	rc = snprintf(fname, MAX_PATH_LEN, "%s/%s", path, name);
```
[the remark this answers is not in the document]
oleksiimoisieiev: +

xen-dom-mgmt/include/domain.h
```diff
@@ -19,6 +20,10 @@ struct xen_domain_iomem {
 	uint64_t nr_mfns;
 };
 
+typedef int (*load_image_bytes_t)(uint8_t *buf, size_t bufsize,
```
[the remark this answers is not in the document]
firscity: Please, add this commit as PR to zephyr-dom0-xt to keep Zephyr DomU runnable

xrun/src/xrun.c
```diff
@@ -0,0 +1,428 @@
...
+
+static const struct json_obj_descr domain_spec_descr[] = {
+	JSON_OBJ_DESCR_PRIM(struct domain_spec, ociVersion, JSON_TOK_STRING),
+	JSON_OBJ_DESCR_OBJECT(struct domain_spec, vm, vm_spec_descr),
+};
+
+int parse_config_json(char *json, size_t json_size, struct domain_spec *domain)
+{
+	int expected_return_code = (1 << ARRAY_SIZE(domain_spec_descr)) - 1;
+	int ret = json_obj_parse(json,
```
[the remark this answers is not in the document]
firscity: Still can see a problem

xrun/src/storage.c
```diff
@@ -0,0 +1,179 @@
...
+		if (rc < 0) {
+			LOG_ERR("FAIL: seek %s: %ld", fname, rc);
+			goto out;
+		}
+	}
+
+	rc = fs_read(&file, buf, size);
+	if (rc < 0) {
+		LOG_ERR("FAIL: read %s: [rd:%ld]", fname, rc);
+		goto out;
```
xakep-amatop, line 52: (cosmetic) unnecessary `goto`

xrun/src/storage.c
```diff
@@ -0,0 +1,179 @@
...
+
+ssize_t read_file(const char *path, const char *name,
+		  char *buf, size_t size, int skip)
+{
+	struct fs_file_t file;
+	ssize_t rc;
+	int ret;
+	char fname[LFS_NAME_MAX];
+
+	__ASSERT(path != NULL && name != NULL, "Invalid input parameters");
```
xakep-amatop, line 25: (question) Do we really need to assert here and in line below?
why we can exit with error ? I thought that code should be reliable

xrun/src/storage.c
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	rc = fs_stat(fname, &dirent);
+	if (rc < 0) {
+		LOG_ERR("FAIL: stat %s: %d", fname, rc);
+		return rc;
+	}
+
+	/* Check if the file exists - if not just write the pattern */
+	if (rc == 0 && dirent.type == FS_DIR_ENTRY_FILE && dirent.size == 0) {
```
xakep-amatop, line 86: (question) is it really impossible to have zero-length file?

xrun/src/storage.c
```diff
@@ -0,0 +1,179 @@
...
+		/* entry.name[0] == 0 means end-of-dir */
+		if (res || entry.name[0] == 0) {
+			if (res < 0) {
+				LOG_ERR("Error reading dir [%d]\n", res);
+			}
+			break;
+		}
+
+		if (entry.type == FS_DIR_ENTRY_DIR) {
+			LOG_DBG("[DIR ] %s\n", entry.name);
```
xakep-amatop, line 126: (question) why do you always add extra new line ? `LOG_DBG` has already added one

xrun/src/storage.c
```diff
@@ -0,0 +1,179 @@
...
+
+LOG_MODULE_REGISTER(storage);
+
+ssize_t read_file(const char *path, const char *name,
+		  char *buf, size_t size, int skip)
+{
+	struct fs_file_t file;
+	ssize_t rc;
+	int ret;
+	char fname[LFS_NAME_MAX];
```
xakep-amatop, line 23: LFS_NAME_MAX only applies to filenames. It mirrors the POSIX [NAME_MAX](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html) constant.

There is no limit on the path length other then the RAM and storage you have on your device. lfs_file_open makes sure to write the file name to disk and doesn't keep it around.

xrun/src/storage.c
xakep-amatop: I have meant that READ flag is useless here, I haven't meant `FS_O_CREATE`...
pls, use `FS_O_WRITE` instead of `FS_O_RDWR`

xakep-amatop (changes requested):

xrun/src/xrun.c
```diff
@@ -0,0 +1,455 @@
...
+		return -EINVAL;
+	}
+
+	config = k_malloc(CONFIG_XRUN_JSON_SIZE_MAX);
+	if (!config) {
+		ret = -ENOMEM;
+		goto err;
+	}
+
+	bytes_read = read_file(bundle, "config.json", config,
```
xakep-amatop, line 329: (question) is it possible to get size of file and compare it with `CONFIG_XRUN_JSON_SIZE_MAX` here and print some warning at least?

xrun/src/xrun.c
```diff
@@ -0,0 +1,455 @@
...
+					CONFIG_XRUN_JSON_SIZE_MAX, 0);
+	if (bytes_read < 0) {
```
xakep-amatop, line 331: (question) is it possible to have `bytes_read` is equal to `0` here?

xrun/src/xrun.c
```diff
@@ -0,0 +1,455 @@
...
+		LOG_ERR("Can't read config.json ret = %ld\n", bytes_read);
```
xakep-amatop, line 332: (cosmetic) pls, remove extra new lines in all files for all logs

xrun/src/xrun.c
```diff
@@ -0,0 +1,455 @@
...
+		ret = bytes_read;
+		goto err_config;
+	}
+
+	ret = parse_config_json(config, bytes_read, &spec);
+	if (ret < 0) {
+		goto err_config;
+	}
+	ret = snprintf(container->kernel_image,
```
xakep-amatop, line 341: (question) Could you please also add checking if we write less bytes than expected?
```
ret < strlen(spec.vm.kernel.path)
```

xrun/src/xrun.c
```diff
@@ -0,0 +1,455 @@
...
+		LOG_ERR("Unable to get kernel path, rc = %d", ret);
+		goto err_config;
+	}
+
+	container->has_dt_image = strlen(spec.vm.hwconfig.devicetree) > 0;
+
+	if (container->has_dt_image) {
+		ret = snprintf(container->dt_image,
+					   MAX_STR_SIZE,
+					   "%s", spec.vm.hwconfig.devicetree);
```
xakep-amatop, line 354: (minor) the same as above

xakep-amatop (changes requested):

Kconfig
```diff
@@ -31,7 +31,14 @@ config XEN_DOMAIN_MANAGEMENT
 	  to create, destroy and manage xen domains.
 
 config XEN_SHELL
+	bool "Enable Xen shell support"
+	help
+	  Enable set of Xen shell support to provide shell for the domains.
+
+config XEN_SHELL_CMDS
```
lorc, line 38: I am moving xen console srv to the separate library `XEN_CONSOLE_SRV`.

If I got this right, `XEN_SHELL` is used only for the shell commands. So, may I ask you to skip this commit? My PR will handle this in the other way around

xrun/src/xrun.c
lorc: @xakep-amatop , GICv3 **may** be backward compatible with GICv2. So on GICv3 machine we **may** support GICv2 for guests.

Anyways, there should be `XEN_DOMCTL_CONFIG_GIC_NATIVE`, I believe.

Kconfig
```diff
@@ -45,3 +45,16 @@ config XEN_VCH
...
 	  Enable vchannel communication library.
+
+config XRUN
+	bool "Enable XRUN support"
+	help
+	  Enable xrun support which allows to run domains based on the runtime spec.
+
+config XRUN_JSON_SIZE_MAX
+	int "Maximum size of the json cni spec"
+	default 8129
```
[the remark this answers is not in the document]
firscity: Yes, but 8129 is not 8K (8129 -> 8192)

xrun/src/xrun.c
```diff
@@ -0,0 +1,455 @@
...
+
+		pos += snprintf(cmdline + pos, size - pos, "%s",
+						spec->vm.kernel.parameters[i]);
+
+		if (i < spec->vm.kernel.params_len - 1) {
+			if (pos + 1 > size) {
+				break;
+			}
+
+			cmdline[pos++] = ' ';
```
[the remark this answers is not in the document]
oleksiimoisieiev: fixed

xrun/src/xrun.c
```diff
@@ -0,0 +1,455 @@
...
+{
+	int i, pos = 0;
+
+	for (i = 0; i < spec->vm.kernel.params_len; i++) {
+		if (pos >= size) {
+			break;
+		}
+
+		pos += snprintf(cmdline + pos, size - pos, "%s",
+						spec->vm.kernel.parameters[i]);
```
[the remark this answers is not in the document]
oleksiimoisieiev: +

xrun/src/xrun.c
oleksiimoisieiev: If we unable to read whole json - we won't parse it. I expect json with correct size in most cases (because app generates json and sets max size) so it will save us one call to storage.

xrun/src/xrun.c
oleksiimoisieiev: yes. why not? If we can parse it.

xrun/src/xrun.c
oleksiimoisieiev: +

xrun/src/xrun.c
oleksiimoisieiev: +

xrun/src/xrun.c
oleksiimoisieiev: +

lorc (commented):
I do understand that time is pressing. So, I am okay with merging this changes as is, as far as you add **BIG FAT** `TODO` regarding object lifetime.

xrun/src/xrun.c
```diff
@@ -0,0 +1,428 @@
...
+	domcfg->dtb_start = container->devicetree;
+	domcfg->dtb_end = container->devicetree + res;
+
+	return 0;
+}
+
+static int parse_cmdline(struct container *container,
+				 char *cmdline, unsigned int size)
+{
+	int i, pos = 0;
```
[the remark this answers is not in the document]
oleksiimoisieiev: I haven't added comments because completely rewritten func where i tried to cover all cases. Sorry, looks like this is an older comment.

xrun/src/xrun.c
```diff
@@ -0,0 +1,481 @@
...
+	unregister_container_id(container_id);
+	return ret;
+}
+
+int xrun_pause(const char *container_id)
+{
+	int ret = 0;
+	struct container *container = get_container(container_id);
+
+	if (!container) {
```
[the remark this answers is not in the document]
oleksiimoisieiev: Added big fat todo.

xrun/src/xrun.c
```diff
@@ -0,0 +1,481 @@
...
+	container->status = PAUSED;
+	return 0;
+}
+
+int xrun_resume(const char *container_id)
+{
+	int ret = 0;
+	struct container *container = get_container(container_id);
+
+	if (!container) {
```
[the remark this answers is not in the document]
oleksiimoisieiev: see above

xrun/src/xrun.c
```diff
@@ -0,0 +1,481 @@
...
+	container->status = RUNNING;
+	return 0;
+}
+
+int xrun_kill(const char *container_id)
+{
+	int ret = 0;
+	struct container *container = get_container(container_id);
+
+	if (!container) {
```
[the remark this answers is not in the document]
oleksiimoisieiev: see above

xrun/src/xrun.c
```diff
@@ -0,0 +1,481 @@
...
+
+int xrun_state(const char *container_id, enum container_status *state)
+{
+	struct container *container = get_container(container_id);
+
+	if (!container) {
+		return -EINVAL;
+	}
+
+	*state = container->status;
```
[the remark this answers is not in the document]
oleksiimoisieiev: see above

xrun/include/storage.h
```diff
@@ -0,0 +1,35 @@
...
+#include <sys/types.h>
+#include <zephyr/types.h>
+
+/*
+ * @brief Read buffer fron file on storage
+ *
+ * @param path - absolute path to the directory
+ * @param name - filename to read buffer from
+ * @param buf - pointer to buffer
+ * @param size - size of the buffer
```
xakep-amatop, line 18: (question) where is description about skip paramr?

xrun/src/storage.c
xakep-amatop: it is really interesting, why can't we have an empty files?..
I have looked on all usages of type and len and I haven't found any checks as you made here

it is true that every directory always has zero length, BUT file also may be with zero length, so I think your check is incorrect here

xrun/src/storage.c
xakep-amatop: > In lfs.c line 2888 we can see the following check: nlen > lfs->name_max which is LFS_NAME_MAX
you should know that `lfs->name_max` is configurable and we have just default config this is the reason why you have seen `LFS_NAME_MAX`

xrun/src/xrun.c
```diff
@@ -0,0 +1,498 @@
...
+	 * ensure that this object does not disappear under
+	 * your feet.
+	 */
+	container = get_container(container_id);
+	if (container) {
+		LOG_ERR("Container %s already exists", container_id);
+		return NULL;
+	}
+
+	k_mutex_lock(&container_lock, K_FOREVER);
```
xakep-amatop, line 173: (note) just move alloc call above mutex lock and you can delete goto and lable at all

xrun/src/xrun.c
xakep-amatop: are you going to parse an empty buffer?

xrun/include/storage.h
oleksiimoisieiev: +

xrun/src/xrun.c
oleksiimoisieiev: +

xrun/src/xrun.c
oleksiimoisieiev: I'm sorry, but I thought I've already answered this question. I'm too tired for discussions. As for me personally - I don't see any problem with empty file or to check every step. I don't think there should be undefined behaviour or unexpected error here. Just tell what to do and I'll do that. That's it.

xrun/src/xrun.c
```diff
@@ -0,0 +1,455 @@
...
+			 unsigned int size)
+{
+	int i, pos = 0;
+
+	for (i = 0; i < spec->vm.kernel.params_len; i++) {
+		if (pos >= size) {
+			break;
+		}
+
+		pos += snprintf(cmdline + pos, size - pos, "%s",
```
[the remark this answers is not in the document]
xakep-amatop: lets go to the sources:
[snprintf](https://elixir.bootlin.com/zephyr/latest/C/ident/snprintf) -> [cbvprintf](https://elixir.bootlin.com/zephyr/latest/C/ident/cbvprintf) -> [z_cbvprintf_impl](https://elixir.bootlin.com/zephyr/latest/C/ident/z_cbvprintf_impl)

and in case when you use function from file [cbprintf_nano.c](https://elixir.bootlin.com/zephyr/latest/source/lib/os/cbprintf_nano.c) it can't ret a negative value at all, but if the function is used from [cbprintf_complete.c](https://elixir.bootlin.com/zephyr/latest/source/lib/os/cbprintf_complete.c) it is possible to ret a negative value from `out` handler, at least you may find appropriate checks of ret code, but `out` handler can't return negative values ([sprintf_out](https://elixir.bootlin.com/zephyr/latest/C/ident/sprintf_out)), at least now...

my point is if function can return negative value you have to handle it in a proper way and don't care about internal realization, because it can be changed

xrun/src/xrun.c
```diff
@@ -0,0 +1,455 @@
...
+	ssize_t bytes_read;
+	char *config;
+	struct domain_spec spec;
+	struct container *container = register_container_id(container_id);
+
+	if (!container) {
+		return -EINVAL;
+	}
+
+	config = k_malloc(CONFIG_XRUN_JSON_SIZE_MAX);
```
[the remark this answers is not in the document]
xakep-amatop: you can avoid using of dynamic allocations and it is preferable way for zephyr code guidelines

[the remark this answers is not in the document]
oleksiimoisieiev: Yes I can but we've discussed it on zehpyr channel.

xrun/src/xrun.c
oleksiimoisieiev: it's json parser who's going to decide whether buffer is correct or not. At this point - it just reads file and doesn't know about it's contents. So why should it tread 0 size ar error?

xakep-amatop: I don't insist on any changes, I just have asked a few questions, because it was really unclear for me, I believe that a little comment helps to understand it more quickly by others too

xrun/src/storage.c
oleksiimoisieiev: +

xrun/src/xrun.c
oleksiimoisieiev: I'm not sure about comment. Let's ask what others are thinking?
@lorc  What do you think if I add the following comment:
"We don't check file size here because at this point the goal is to read the file. It's parser who will have to make a decision, whether file contents are ok or not".

xrun/include/storage.h
```diff
@@ -12,25 +12,23 @@
...
- * @param name - filename to read buffer from
+ * @param fpath - absolute path to the file
  * @param buf - pointer to buffer
  * @param size - size of the buffer
  * @param skip - skip first n bytes before start reading
  *
  * @return - 0 on success and errno on error
  */
-ssize_t xrun_read_file(const char *path, const char *name, char *buf,
+ssize_t xrun_read_file(const char *path, char *buf,
```
xakep-amatop, line 22: (cosmetic) different names of first parameter here and in description

xrun/src/xrun.c
```diff
@@ -361,15 +361,40 @@ static int generate_cmdline(struct domain_spec *spec, char **cmdline)
...
+static ssize_t get_fpath_size(const char *path, const char *name)
+{
+	size_t target_size = 0;
+
+	if (!path || !name) {
+		LOG_ERR("Invalid input parameters");
+		return -EINVAL;
+	}
+
+	if (strlen(path) == 0 || strlen(name) == 0) {
```
xakep-amatop, line 373: (note) you can add extra two variables and avoid call of `strlen` second time below

xrun/src/xrun.c
```diff
@@ -379,14 +404,37 @@ int xrun_run(const char *bundle, int console_socket, const char *container_id)
...
+		k_free(fpath);
+		goto err_config;
+	}
+
+	bytes_read = xrun_read_file(fpath, config,
 				    CONFIG_XRUN_JSON_SIZE_MAX, 0);
 	if (bytes_read < 0) {
 		LOG_ERR("Can't read config.json ret = %ld", bytes_read);
 		ret = bytes_read;
+		k_free(fpath);
```
xakep-amatop, line 432: (note) you can move this mem free rigth after a call of `xrun_read_file` and simplify the code

xakep-amatop:

xrun/src/xrun.c
oleksiimoisieiev: I've added after check because this approach seems to me more clear. And it's harder to miss k_free during refactoring

xrun/src/xrun.c
```diff
@@ -0,0 +1,508 @@
...
+	}
+
+	container = (struct container *)k_malloc(sizeof(*container));
+	if (!container) {
+		return NULL;
+	}
+
+	k_mutex_lock(&container_lock, K_FOREVER);
+
+	strncpy(container->container_id, container_id, CONTAINER_NAME_SIZE);
```
firscity, line 182: The only `sys_slist_append()` should be protected under the mutex.

firscity:

lorc:

PR #18, opened 2023-03-20
xenstore-srv: refactor watch notification to avoid domain list reference

firscity:
This commit changes Xenstore watches logic, to prevent xenstore-srv from using global domain list during watcher notification. Referencing to domain list without proper sycnronization between xen-dom-mgmt and xenstore-srv is risky and actually was not needed.

To notify watcher domain about file changes, we need to add pending event to global pevent list and notify domain with xb_sem. Previously, watch event holded only domid, which needed a domid_to_domain() calls for notification. Now we changed domid inside this structs to domain pointers to make direct notification possible and also fixed possible watches leaks during domain destroy. In this implementation all watches and pending events will be removed during domain destroy.

### DIFF
--- a/xenstore-srv/include/xenstore_srv.h
+++ b/xenstore-srv/include/xenstore_srv.h
@@ -21,7 +21,7 @@ struct xs_entry {
 struct watch_entry {
 	char *key;
 	char *token;
-	int domid;
+	struct xen_domain *domain;
 	bool is_relative;
 
 	sys_dnode_t node;
@@ -29,7 +29,7 @@ struct watch_entry {
 
 struct pending_watch_event_entry {
 	char *key;
-	int domid;
+	struct xen_domain *domain;
 
 	sys_dnode_t node;
 };
--- a/xenstore-srv/src/xenstore_srv.c
+++ b/xenstore-srv/src/xenstore_srv.c
@@ -371,43 +371,68 @@ int xss_set_perm(const char *path, domid_t domid, enum xs_perm perm)
 	return 0;
 }
 
-/*
- * TODO: this function should not be called from anywhere except event channel
- * handler. Need to understand why notify_sibling_domains() works in such way
- * and fix it. Declaration is needed to avoid compiler warnings during build.
- */
-void xs_evtchn_cb(void *priv);
-
-//TODO: header, moveto dom0.c
-void notify_sibling_domains(uint32_t *sdom_list, size_t len)
+static void notify_watchers(const char *path, uint32_t caller_domid)
 {
-	struct xen_domain *iter;
+	struct watch_entry *iter;
+	struct pending_watch_event_entry *pentry;
 
-	for (size_t sdi=0;sdom_list[sdi];++sdi)
-	{
-		iter = domid_to_domain(sdom_list[sdi]);
-
-		if (iter)
-		{
-			xs_evtchn_cb((void*)iter);
+	k_mutex_lock(&wel_mutex, K_FOREVER);
+	SYS_DLIST_FOR_EACH_CONTAINER(&watch_entry_list, iter, node) {
+		if (iter->domain->domid == caller_domid ||
+		    strncmp(iter->key, path, strlen(iter->key))) {
+			continue;
 		}
+
+		pentry = k_malloc(sizeof(*pentry));
+		if (!pentry) {
+			goto pentry_fail;
+		}
+
+		pentry->key = k_malloc(strlen(path) + 1);
+		if (!pentry->key) {
+			goto pkey_fail;
+		}
+
+		strcpy(pentry->key, path);
+		pentry->domain = iter->domain;
+
+		sys_dnode_init(&pentry->node);
+		k_mutex_lock(&pfl_mutex, K_FOREVER);
+		sys_dlist_append(&pending_watch_event_list,
+				 &pentry->node);
+		k_mutex_unlock(&pfl_mutex);
+
+		/* Wake watcher thread up */
+		k_sem_give(&iter->domain->xb_sem);
+
 	}
+	k_mutex_unlock(&wel_mutex);
+
+	return;
+
+pkey_fail:
+	k_free(pentry);
+pentry_fail:
+	k_mutex_unlock(&wel_mutex);
+	LOG_WRN("Failed to notify Domain#%d about path %s, no memory",
+		iter->domain->domid, path);
 }
 
 void _handle_write(struct xen_domain *domain, uint32_t id, uint32_t msg_type, char *payload,
 		   uint32_t len)
 {
+	char localpath[] = "/";
+	char path[STRING_LENGTH_MAX];
+	char *data;
 	uint32_t data_offset = strlen(payload) + 1;
-	char *data = payload + data_offset;
+
+	data = payload + data_offset;
 
 	if (len < data_offset) {
 		LOG_ERR("Data size mismatch");
 		send_errno(domain, id, EINVAL);
 		return;
 	}
-
-	const char localpath[] = "/";
-	char path[STRING_LENGTH_MAX];
 
 	if (memcmp(payload, localpath, strlen(localpath)) == 0) {
 		memcpy(path, payload, data_offset);
@@ -420,43 +445,7 @@ void _handle_write(struct xen_domain *domain, uint32_t id, uint32_t msg_type, ch
 
 	send_reply(domain, id, msg_type, "OK");
 
-	struct watch_entry *iter;
-
-	uint32_t sibling_domains[DOM_MAX] = {0};
-	k_mutex_lock(&wel_mutex, K_FOREVER);
-	SYS_DLIST_FOR_EACH_CONTAINER (&watch_entry_list, iter, node) {
-		uint32_t iklen = strlen(iter->key);
-		if (memcmp(iter->key, path, iklen) == 0) {
-			struct pending_watch_event_entry *entry;
-			entry = k_malloc(sizeof(*entry));
-			size_t keyl = strlen(path) + 1;
-			entry->key = k_malloc(keyl);
-			memcpy(entry->key, path, keyl);
-			entry->key[keyl - 1] = 0;
-			entry->domid = iter->domid;
-
-			if (iter->domid != domain->domid)
-			{
-				for (size_t sdi=0;sdi<DOM_MAX;++sdi)
-				{
-					if (sibling_domains[sdi] == 0 || sibling_domains[sdi] == iter->domid)
-					{
-						sibling_domains[sdi] = iter->domid;
-						break;
-					}
-				}
-			}
-
-			k_mutex_lock(&pfl_mutex, K_FOREVER);
-			sys_dnode_init(&entry->node);
-			sys_dlist_append(&pending_watch_event_list, &entry->node);
-			k_mutex_unlock(&pfl_mutex);
-		}
-	}
-
-	k_mutex_unlock(&wel_mutex);
-
-	notify_sibling_domains(sibling_domains, DOM_MAX);
+	notify_watchers(path, domain->domid);
 }
 
 void handle_write(struct xen_domain *domain, uint32_t id, char *payload, uint32_t len)
@@ -472,20 +461,24 @@ void handle_mkdir(struct xen_domain *domain, uint32_t id, char *payload, uint32_
 void process_pending_watch_events(struct xen_domain *domain, uint32_t id)
 {
 	struct pending_watch_event_entry *iter, *next;
-	k_mutex_lock(&pfl_mutex, K_FOREVER);
 
+	k_mutex_lock(&pfl_mutex, K_FOREVER);
 	SYS_DLIST_FOR_EACH_CONTAINER_SAFE (&pending_watch_event_list, iter, next, node) {
-		if (domain->domid == iter->domid && domain->running_transaction == 0 && fire_watcher(domain, id, iter->key)) {
-			if (domain->pending_stop_transaction == true && domain->stop_transaction_id == 0)
-			{
-				continue;
+		/* TODO: check and simplify this if statements if possible */
+		if (domain == iter->domain) {
+			if (domain->running_transaction == 0 &&
+			    fire_watcher(domain, id, iter->key)) {
+				if (domain->pending_stop_transaction == true &&
+				    domain->stop_transaction_id == 0) {
+					continue;
+				}
+
+				k_free(iter->key);
+				sys_dlist_remove(&iter->node);
+				k_free(iter);
 			}
-			k_free(iter->key);
-			sys_dlist_remove(&iter->node);
-			k_free(iter);
 		}
 	}
-
 	k_mutex_unlock(&pfl_mutex);
 }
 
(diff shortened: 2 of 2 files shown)


xenstore-srv/include/xenstore_srv.h
```diff
@@ -21,15 +21,15 @@ struct xs_entry {
 struct watch_entry {
```
oleksiimoisieiev, line 21: Please run checkpatch. I've got 5 warnings.

xenstore-srv/src/xenstore_srv.c
```diff
@@ -371,44 +371,68 @@ int xss_set_perm(const char *path, domid_t domid, enum xs_perm perm)
...
+static void notify_watchers(const char *path, uint32_t caller_domid)
 {
-	struct xen_domain *iter;
+	struct watch_entry *iter;
 
-	for (size_t sdi=0;sdom_list[sdi];++sdi)
-	{
-		iter = domid_to_domain(sdom_list[sdi]);
+	k_mutex_lock(&wel_mutex, K_FOREVER);
+	SYS_DLIST_FOR_EACH_CONTAINER(&watch_entry_list, iter, node) {
```
oleksiimoisieiev, line 379: As I can see - all other iterations with watch_entry_list are done using FOR_EACH_ENTRY_SAFE, I looked through your part and don't see any problem with detached iterators but I could miss something. You don't necessary need to fix this. Just a tip to consider.

oleksiimoisieiev (changes requested):

xenstore-srv/src/xenstore_srv.c
```diff
@@ -371,44 +371,68 @@ int xss_set_perm(const char *path, domid_t domid, enum xs_perm perm)
...
+				return;
+			}
+
+			keyl = strlen(path) + 1;
+			pentry->key = k_malloc(keyl);
+			if (!pentry->key) {
+				k_mutex_unlock(&wel_mutex);
+
+				k_free(pentry);
+				return;
```
lorc, line 400: The same comment about warning.

xenstore-srv/src/xenstore_srv.c
```diff
@@ -371,44 +371,68 @@ int xss_set_perm(const char *path, domid_t domid, enum xs_perm perm)
...
+			}
+
+			memcpy(pentry->key, path, keyl);
+			pentry->key[keyl - 1] = 0;
```
lorc, line 404: Again, why not `strcpy`?

xenstore-srv/src/xenstore_srv.c
```diff
@@ -371,44 +371,68 @@ int xss_set_perm(const char *path, domid_t domid, enum xs_perm perm)
...
 		}
 	}
+
+	k_mutex_unlock(&wel_mutex);
 }
 
 void _handle_write(struct xen_domain *domain, uint32_t id, uint32_t msg_type, char *payload,
 		   uint32_t len)
 {
+	char localpath[] = "/";
```
lorc, line 423: It is good to do some cleanup while reworking those parts. But you need to mention this in the commit message, at least.

Ideally - it should go to another commit.

xenstore-srv/src/xenstore_srv.c
```diff
@@ -472,20 +460,24 @@ void handle_mkdir(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
-	k_mutex_lock(&pfl_mutex, K_FOREVER);
 
+	k_mutex_lock(&pfl_mutex, K_FOREVER);
 	SYS_DLIST_FOR_EACH_CONTAINER_SAFE (&pending_watch_event_list, iter, next, node) {
-		if (domain->domid == iter->domid && domain->running_transaction == 0 && fire_watcher(domain, id, iter->key)) {
-			if (domain->pending_stop_transaction == true && domain->stop_transaction_id == 0)
-			{
-				continue;
+		/* TODO: check and simplify this if statements if possible */
+		if (domain->domid == iter->domain->domid) {
```
lorc, line 467: You can compare `domain` to `iter->domain`, no need to check IDs.

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,54 +603,91 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
-		entry->token = k_malloc(len - plen);
-		memcpy(entry->token, token, len - plen);
-		entry->domid = domain->domid;
+	wentry = key_to_watcher(path, true, token);
+	full_plen = strlen(path) + 1;
+
+	if (!wentry) {
+		wentry = k_malloc(sizeof(*wentry));
+		if (!wentry) {
+			return;
```
lorc, line 634: It would be nice to return some error to the calling domain.

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,54 +603,91 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
+		}
+
+		wentry->key = k_malloc(full_plen);
+		if (!wentry->key) {
+			k_free(wentry);
+			return;
```
lorc, line 640: It would be nice to return some error to the calling domain.

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,54 +603,91 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
+		}
+
+		wentry->token = k_malloc(len - plen);
+		if (!wentry->token) {
+			k_free(wentry->key);
+			k_free(wentry);
+			return;
```
lorc, line 647: It would be nice to return some error to the calling domain.

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,54 +603,91 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
-
 	if (key_to_entry(path)) {
-		struct pending_watch_event_entry *entry;
-		entry = k_malloc(sizeof(*entry));
-		entry->key = k_malloc(lpath);
-		memcpy(entry->key, path, lpath);
-		entry->domid = domain->domid;
+		pentry = k_malloc(sizeof(*pentry));
+		if (!pentry) {
+			/* We can't notify domain so leave it without first WATCH_EVENT */
```
lorc, line 668: I believe, this is not the worst in this situation... At least we should output warning to the log.

xenstore-srv/src/xenstore_srv.c
```diff
@@ -682,7 +711,7 @@ void handle_unwatch(struct xen_domain *domain, uint32_t id, char *payload, uint3
 	struct watch_entry *entry = key_to_watcher(path, true, token);
 
 	if (entry) {
-		if (entry->domid == domain->domid) {
+		if (entry->domain->domid == domain->domid) {
```
lorc, line 714: You can compare two domain pointers there.

lorc:
I do understand that the original code is messy. But it will be good to have code cleanup and logic changes in a different commits. If this is possible, of course.

xenstore-srv/src/xenstore_srv.c
```diff
@@ -793,6 +822,31 @@ int stop_domain_stored(struct xen_domain *domain)
...
+	struct pending_watch_event_entry *pwe_iter, *pwe_next;
+
+	k_mutex_lock(&wel_mutex, K_FOREVER);
+	SYS_DLIST_FOR_EACH_CONTAINER_SAFE(&watch_entry_list, iter, next, node) {
+		if (iter->domain == domain) {
+			remove_watch_entry(iter);
+		}
+	}
+	k_mutex_unlock(&wel_mutex);
+
```
LKomaryanskiy, line 857: cosmetic: please, remove two newlines

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,54 +603,91 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
-	size_t lpath = strlen(path) + 1;
-
-	if (!entry) {
-		entry = k_malloc(sizeof(*entry));
-		entry->key = k_malloc(lpath);
-		memcpy(entry->key, path, lpath);
-		entry->token = k_malloc(len - plen);
-		memcpy(entry->token, token, len - plen);
-		entry->domid = domain->domid;
+	wentry = key_to_watcher(path, true, token);
```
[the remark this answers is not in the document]
LKomaryanskiy: Also, key_to_watcher and fire_watcher make access to watch_entry_list without wel_mutex lock. Could you please check if is it ok?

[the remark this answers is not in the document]
LKomaryanskiy: BTW, key_to_watcher returns pointer to wentry but I assume, that it can be situation, when after that can be called function that will free wentry via remove_watch_entry, so assigning wentry->is_relative will lead to use after free issue.

xenstore-srv/src/xenstore_srv.c
firscity: Thanks, done.

xenstore-srv/src/xenstore_srv.c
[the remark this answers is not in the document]
firscity: > As I see, this function is called in two places, and in both cases second parameter (complete) is true. Maybe is it better to remove it at all?

I think it is somehow related to watches to directories. I will check it in further PR, xenstore is not fully implemented now.

> Also, key_to_watcher and fire_watcher take access to watch_entry_list without wel_mutex lock. Could you please check if is it ok?

> BTW, key_to_watcher returns pointer to wentry but I assume, that it can be situation, when after that can be called function that will free wentry via remove_watch_entry, so assigning to wentry->is_relative can situably lead to use after free memory issue.

No, probably it is not. Fixing require some time, I'm currently working on it. I fixed only call to `key_to_watcher` in my PR, please check.

LKomaryanskiy (approved):

lorc:

xenstore-srv/src/xenstore_srv.c
```diff
@@ -371,44 +371,70 @@ int xss_set_perm(const char *path, domid_t domid, enum xs_perm perm)
...
+static void notify_watchers(const char *path, uint32_t caller_domid)
 {
-	struct xen_domain *iter;
+	struct watch_entry *iter;
+	struct pending_watch_event_entry *pentry;
 
-	for (size_t sdi=0;sdom_list[sdi];++sdi)
-	{
-		iter = domid_to_domain(sdom_list[sdi]);
+	k_mutex_lock(&wel_mutex, K_FOREVER);
```
xakep-amatop, line 379: (question major) Why didn't you check status here?
```
 * @retval 0 Mutex locked.
 * @retval -EBUSY Returned without waiting.
 * @retval -EAGAIN Waiting period timed out.
 */
__syscall int k_mutex_lock(struct k_mutex *mutex, k_timeout_t timeout);
```

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,55 +605,116 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
+	path_is_relative = !!(memcmp(payload, localpath, strlen(localpath)));
+
+	/*
+	 * Path and token come inside payload char * and are separated
+	 * with '\0', so we can find path len with strnlen here.
+	 */
+	plen = strnlen(payload, len) + 1;
+
+	if (path_is_relative) {
+		snprintf(path, STRING_LENGTH_MAX, "/local/domain/%d/%s",
```
xakep-amatop, line 622: (note) this function can return negative error code  or 0

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,55 +605,116 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
+			 domain->domid, payload);
```
xakep-amatop, line 623: (question) could you please add check if you have enough space inside buffer ?

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,55 +605,116 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
 	} else {
 		memcpy(path, payload, plen);
 	}
+	full_plen = strlen(path) + 1;
 
+	/* Extract token value from payload (between 'path' and end '\0') */
 	memcpy(token, payload + plen, len - plen);
```
xakep-amatop, line 630: (major) could you pls add extra checking if we have enough space inside `token`:
```
if (sizeof(token) < len - plen)
```
**possible stack corruption here**

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,55 +605,116 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
-		entry->key = k_malloc(lpath);
-		memcpy(entry->key, path, lpath);
-		entry->token = k_malloc(len - plen);
-		memcpy(entry->token, token, len - plen);
-		entry->domid = domain->domid;
-		k_mutex_lock(&wel_mutex, K_FOREVER);
-		sys_dnode_init(&entry->node);
-		sys_dlist_append(&watch_entry_list, &entry->node);
+	k_mutex_lock(&wel_mutex, K_FOREVER);
+	wentry = key_to_watcher(path, true, token);
```
xakep-amatop, line 633: (note/not your issue) as far as I see this function is used only inside this file, but it hasn't `static` keyword

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,55 +605,116 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
+		}
+
+		wentry->token = k_malloc(len - plen);
+		if (!wentry->token) {
+			goto wtoken_fail;
+		}
 
-	entry->is_relative = is_relative;
+		memcpy(wentry->key, path, full_plen);
+		memcpy(wentry->token, token, len - plen);
```
xakep-amatop, line 659: (note) you limit buffer on stack, but don't limit buffers during allocations of `key` and `token`
as for me it will be better to put key and token as an arrays inside `wentry` and use only one alloc for `wentry`

moreover, `len - plen` can be bigger than size of `token`

how much memory did you save and how much CPU cycles waste for a few allocations?

xenstore-srv/src/xenstore_srv.c
xakep-amatop, line 622: (note) better to use sizeof here instead of `STRING_LENGTH_MAX` in this case if somebody changes len of path array you don't need to change it in all places

xenstore-srv/src/xenstore_srv.c
```diff
@@ -371,44 +371,70 @@ int xss_set_perm(const char *path, domid_t domid, enum xs_perm perm)
...
-	struct xen_domain *iter;
+	struct watch_entry *iter;
+	struct pending_watch_event_entry *pentry;
 
-	for (size_t sdi=0;sdom_list[sdi];++sdi)
-	{
-		iter = domid_to_domain(sdom_list[sdi]);
+	k_mutex_lock(&wel_mutex, K_FOREVER);
+	SYS_DLIST_FOR_EACH_CONTAINER(&watch_entry_list, iter, node) {
+		if (strncmp(iter->key, path, strlen(iter->key)) == 0) {
```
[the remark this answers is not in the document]
firscity: Done

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,55 +605,116 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
+	 */
+	plen = strnlen(payload, len) + 1;
+
+	if (path_is_relative) {
+		snprintf(path, STRING_LENGTH_MAX, "/local/domain/%d/%s",
+			 domain->domid, payload);
 	} else {
 		memcpy(path, payload, plen);
 	}
+	full_plen = strlen(path) + 1;
```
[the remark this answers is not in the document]
firscity: Done

firscity (commented):
Thanks for review, please check updates.

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,55 +604,131 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
+	 */
+	plen = strnlen(payload, len) + 1;
+	if (plen > STRING_LENGTH_MAX) {
+		goto path_fail;
+	}
+
+	if (path_is_relative) {
+		full_plen = snprintf(path, sizeof(path), "/local/domain/%d/%s",
+			 domain->domid, payload);
+		if (full_plen < 0) {
```
xakep-amatop, line 626: (note) `full_plen` can be equal to zero

xakep-amatop (approved):

xenstore-srv/src/xenstore_srv.c
```diff
@@ -611,55 +604,131 @@ void handle_watch(struct xen_domain *domain, uint32_t id, char *payload, uint32_
...
-		sys_dlist_append(&watch_entry_list, &entry->node);
+	k_mutex_lock(&wel_mutex, K_FOREVER);
+	wentry = key_to_watcher(path, true, token);
+
+	if (wentry) {
+		/* Same watch, different path form */
+		wentry->is_relative = path_is_relative;
 		k_mutex_unlock(&wel_mutex);
-	}
+	} else {
```
oleksiimoisieiev, line 647: k_mutex_unlock in both conditions. it can be placed before if

xenstore-srv/src/xenstore_srv.c
oleksiimoisieiev: Hm. snprintf will return at least strlen("/local/domain/"). it can return 0 only if sizeof(path) = 0, which is unlikely. As for me - check for 0 here is redundand

xenstore-srv/src/xenstore_srv.c
oleksiimoisieiev: Ok.

PR #21, opened 2023-03-24
Rework Xen Console driver

lorc:
Now you can use `xu console_attach -d NNN` command to attach to a domain console.

### DIFF
--- a/xen-console-srv/src/xen_console.c
+++ b/xen-console-srv/src/xen_console.c
@@ -241,13 +241,16 @@ int xen_init_domain_console(struct xen_domain *domain)
 
 	rc = hvm_set_parameter(HVM_PARAM_CONSOLE_EVTCHN, domain->domid,
 			       console->evtchn);
-
 	if (rc) {
 		LOG_ERR("Failed to set domain console evtchn param (rc=%d)", rc);
-		goto err_free;
+		goto err_unbind;
 	}
 
 	return 0;
+
+err_unbind:
+	unbind_event_channel(console->local_evtchn);
+	evtchn_close(console->local_evtchn);
 
 err_free:
 	k_free(console->int_buf);
--- a/xen-console-srv/src/xen_console.c
+++ b/xen-console-srv/src/xen_console.c
@@ -22,10 +22,12 @@
 
 LOG_MODULE_REGISTER(xen_domain_console);
 
-#define XSS_CONSOLE_STACK_SIZE_PER_DOM 8192
-K_KERNEL_STACK_DEFINE(read_thrd_stack, XSS_CONSOLE_STACK_SIZE_PER_DOM * DOM_MAX);
 static size_t stack_slots[DOM_MAX] = { 0 };
+/* One page is enough for anyone */
+#define XEN_CONSOLE_STACK_SIZE		4096
 
+static K_THREAD_STACK_ARRAY_DEFINE(read_thrd_stack, DOM_MAX,
+				   XEN_CONSOLE_STACK_SIZE);
 /*
  * Need to read from OUT ring in dom0, domU writes logs there
  * TODO: place this in separate driver
@@ -139,8 +141,10 @@ int start_domain_console(struct xen_domain *domain)
 	k_sem_init(&domain->console_sem, 1, 1);
 	domain->console_thrd_stop = false;
 	domain->console_tid =
-		k_thread_create(&domain->console_thrd, read_thrd_stack + XSS_CONSOLE_STACK_SIZE_PER_DOM * slot,
-				K_KERNEL_STACK_SIZEOF(read_thrd_stack) / DOM_MAX, console_read_thrd, domain,
+		k_thread_create(&domain->console_thrd,
+				read_thrd_stack[slot],
+				XEN_CONSOLE_STACK_SIZE,
+				console_read_thrd, domain,
 				NULL, NULL, 7, 0, K_NO_WAIT);
 
 	return 0;
--- a/xen-console-srv/src/xen_console.c
+++ b/xen-console-srv/src/xen_console.c
@@ -13,6 +13,7 @@
 #include <zephyr/init.h>
 #include <zephyr/kernel.h>
 #include <zephyr/logging/log.h>
+#include <zephyr/shell/shell.h>
 #include <string.h>
 #include <stdio.h>
 
@@ -27,15 +28,21 @@ LOG_MODULE_REGISTER(xen_domain_console);
 /* Need low prio to make sure that guest does not lock up us in reader thread */
 #define XEN_CONSOLE_PRIO		14
 #define EXT_THREAD_STOP_BIT		0
+#define INT_THREAD_STOP_BIT		1
 
 /* Size is chosen based on educated guess. It should be power of two. */
 #define XEN_CONSOLE_BUFFER_SZ		8192
 
+#define ESCAPE_CHARACTER		0x1d /* CTR+] */
+
 static K_THREAD_STACK_ARRAY_DEFINE(read_thrd_stack, DOM_MAX,
 				   XEN_CONSOLE_STACK_SIZE);
+static K_THREAD_STACK_DEFINE(display_thrd_stack, XEN_CONSOLE_STACK_SIZE);
 
 static uint32_t used_threads;
 static K_MUTEX_DEFINE(global_console_lock);
+/* There we store pointer to an attached console */
+static struct xen_domain_console *current_console;
 
 BUILD_ASSERT(sizeof(used_threads) * CHAR_BIT >= DOM_MAX);
 BUILD_ASSERT(XEN_CONSOLE_BUFFER_SZ &&
@@ -92,6 +99,42 @@ static void console_feed_int_ring(struct xen_domain_console *console, char ch)
 	}
 }
 
+/* Write data to DomU
+ * Please note that ring buffers named in accordance to DomU point of view:
+ * intf->in is DomU input and our output;
+ */
+static void write_to_ext_ring(struct xencons_interface *intf,
+			      char *data, int len)
+{
+	XENCONS_RING_IDX cons = intf->in_cons;
+	XENCONS_RING_IDX prod = intf->in_prod;
+	XENCONS_RING_IDX idx = 0;
+	size_t free_space;
+
+	compiler_barrier();
+	if ((prod - cons) > sizeof(intf->in)) {
+		LOG_WRN("Invalid state of console input ring. Resetting.");
+		intf->in_prod = cons;
+		return;
+	}
+
+	free_space = sizeof(intf->in) - (prod-cons);
+	if (free_space < len) {
+		len = free_space;
+		/* We can't block and we can't even print a warning */
+	}
+	while (len) {
+		idx = MASK_XENCONS_IDX(prod, intf->out);
+		intf->in[idx] = *data;
+		prod++;
+		data++;
+		len--;
+	}
+
+	compiler_barrier();
+	intf->in_prod = prod;
+}
+
 /* Read from domU ring buffer into a local ring buffer.
  * Please note that ring buffers named in accordance to DomU point of view:
  * intf->out is DomU output and our input;
@@ -143,6 +186,8 @@ static void console_read_thrd(void *con, void *p2, void *p3)
 		while (read_from_ext_ring(console->intf, console))
 			;
 		k_mutex_unlock(&console->lock);
+		/* Notify display thread (if any) */
+		k_sem_give(&console->int_sem);
 	}
 }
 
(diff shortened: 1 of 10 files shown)


xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -909,9 +881,9 @@ int domain_destroy(uint32_t domid)
 
 	stop_domain_stored(domain);
 	/* TODO: do this on console destroying */
-	stop_domain_console(domain);
+	xen_stop_domain_console(domain);
```
xakep-amatop, line 884: (question) should you add checking of error code here?

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -9,6 +9,7 @@
 #include <stdlib.h>
 
 #include <xen_dom_mgmt.h>
+#include <xen_console.h>
 
 LOG_MODULE_DECLARE(xen_shell);
```
xakep-amatop, line 14: (major) pls, change this macro to `LOG_MODULE_REGISTER`, because we don't have register macro anymore after deleting of `xen_shell.c` file

xen-console-srv/src/xen_console.c
```diff
@@ -0,0 +1,456 @@
...
+BUILD_ASSERT((XEN_CONSOLE_BUFFER_SZ & (XEN_CONSOLE_BUFFER_SZ - 1)) == 0);
+
+/* Allocate one stack for external reader thread */
+static int get_stack_idx(void)
+{
+	int i, ret = -ENOSPC;
+
+	k_mutex_lock(&global_console_lock, K_FOREVER);
+
+	for (i = 0; i < DOM_MAX; i++)
```
xakep-amatop, line 57: (cosmetic) pls, add curly braces

xen-console-srv/src/xen_console.c
```diff
@@ -0,0 +1,456 @@
...
+/* Size is chosen based on educated guess. It should be power of two. */
+#define XEN_CONSOLE_BUFFER_SZ		8192
+
+#define ESCAPE_CHARACTER		0x1d /* CTR+] */
+
+static K_THREAD_STACK_ARRAY_DEFINE(read_thrd_stack, DOM_MAX,
+				   XEN_CONSOLE_STACK_SIZE);
+static K_THREAD_STACK_DEFINE(display_thrd_stack, XEN_CONSOLE_STACK_SIZE);
+
+static uint32_t used_threads;
```
xakep-amatop, line 42: (question) Is it enough for us to have 32-threads?

xen-console-srv/src/xen_console.c
xakep-amatop, line 57: (note) looks like you can do it faster:
```
uint32_t inv = ~used_threads;
if (inv) {
    ret = __builtin_ffs(inv) - 1;
    used_threads |= BIT(ret);
    LOG_DBG("Allocated stack with index %d\n", i);
}
```

xen-console-srv/src/xen_console.c
```diff
@@ -0,0 +1,456 @@
...
+static K_MUTEX_DEFINE(global_console_lock);
+/* There we store pointer to an attached console */
+static struct xen_domain_console *current_console;
+
+BUILD_ASSERT(sizeof(used_threads) * CHAR_BIT >= DOM_MAX);
+BUILD_ASSERT((XEN_CONSOLE_BUFFER_SZ & (XEN_CONSOLE_BUFFER_SZ - 1)) == 0);
```
xakep-amatop, line 48: (note) it isn't work for case when `XEN_CONSOLE_BUFFER_SZ` is equal to `0`, could you please add extra check for it?

xen-console-srv/src/xen_console.c
```diff
@@ -123,26 +161,23 @@ int init_domain_console(struct xen_domain *domain)
...
 	}
 
-	for (; slot < DOM_MAX && stack_slots[slot] != 0; ++slot);
-
-	if (slot >= DOM_MAX) {
-		LOG_ERR("Unable to find memory for console stack (%zu >= MAX:%d)", slot, DOM_MAX);
-		return 1;
+	rc = get_stack_idx();
+	if (rc < 0) {
+		return rc;
```
LKomaryanskiy, line 277: (question) Should we add some error message here?

xen-dom-mgmt/src/xen-dom-mgmt.c
LKomaryanskiy: Sorry for answering not for me question, but I did this in my PR:
https://github.com/xen-troops/zephyr-xenlib/pull/13/commits/15033fb61edf5764dc898a6ef907d92cacbe8e46
I will rebase it after @lorc PR will be merged.

xen-console-srv/src/xen_console.c
```diff
@@ -221,6 +248,8 @@ int xen_stop_domain_console(struct xen_domain *domain)
 
 	console = &domain->console;
 
+	k_free(console->int_buf);
```
LKomaryanskiy, line 305: As I see, stopping the thread executed after this line and perhaps at this time gap console_feed_int_ring that uses console->int_buf can be invoked. Seems like can it lead to use after free issue, or not?

xen-console-srv/src/xen_console.c
```diff
@@ -0,0 +1,456 @@
...
+	__ASSERT_NO_MSG(console != NULL);
+
+	if (len == 1 && data[0] == ESCAPE_CHARACTER) {
+		/* Detach console */
+		atomic_set_bit(&console->stop_thrd, INT_THREAD_STOP_BIT);
+		k_sem_give(&console->int_sem);
+		shell_set_bypass(shell, NULL);
+	}
+
+	write_to_ext_ring(console->intf, data, len);
```
xakep-amatop, line 413: (question major) why didn't you protect writes with some locks/mutexes?
as far as I see, you protected reads, I think that can be some concurrency problems on multicore systems related to `intf->in_prod`

xen-console-srv/src/xen_console.c
```diff
@@ -0,0 +1,456 @@
...
+		}
+
+		while (console->int_cons < console->int_prod) {
+			read = 0;
+			memset(buf, 0, sizeof(buf));
+
+			/* TODO: There is a room for optimization.... */
+			while (console->int_cons < console->int_prod &&
+			       read < sizeof(buf) - 1) {
+				size_t buf_pos = (console->int_cons++) %
```
xakep-amatop, line 377: (note) fill free to use `&` instead of `%`, because `XEN_CONSOLE_BUFFER_SZ` is power of two

xen-console-srv/src/xen_console.c
```diff
@@ -0,0 +1,456 @@
...
+	if (current_console) {
+		/* Actually, this should never happen */
+		shell_error(shell, "Shell is already attached to console");
+		k_mutex_unlock(&global_console_lock);
+		return -EEXIST;
+	}
+	current_console = console;
+	k_mutex_unlock(&global_console_lock);
+
+	if (!console->first_attach) {
```
xakep-amatop, line 438: (question) is it better to use here atomic cmp xchg or even console lock ?

xakep-amatop (changes requested):

xen-console-srv/src/xen_console.c
lorc: `DOM_MAX` is `4` right now, so no problems right now. We can easily extend this variable to `uint64_t` if there will be need. And if we ever need to run more than 64 DomUs, we will have more serious problems, than this.

Also, note that I'm adding `BUILD_ASSERT` just in case.

xen-console-srv/src/xen_console.c
lorc: Good idea, thanks.

xen-console-srv/src/xen_console.c
```diff
@@ -0,0 +1,456 @@
...
+	 * variable to true, so then attach handler can reset it to false
+	 */
+	console->first_attach = true;
+
+	rc = bind_interdomain_event_channel(domain->domid,
+					    console->evtchn,
+					    evtchn_callback, console);
+
+	if (rc < 0) {
+		goto err_free;
```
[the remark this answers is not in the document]
lorc: No, I should not. Why?

xen-console-srv/src/xen_console.c
lorc: Ahh, good catch. Thank you. I need to revisit this part.

xen-console-srv/src/xen_console.c
```diff
@@ -0,0 +1,456 @@
...
+#ifdef CONFIG_XEN_SHELL
+
+static void console_display_thrd(void *p1, void *p2, void *p3)
+{
+	ARG_UNUSED(p2);
+	ARG_UNUSED(p3);
+	struct xen_domain_console *console = p1;
+	const struct shell *shell = p2;
+	/* Buffer input a little */
+	char buf[32];
```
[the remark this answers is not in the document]
lorc: There is the comment right before this. I see no profit in moving this into define, as it is used exactly in one place and is not this important. 

But,  if you insist, I can add a separate define like `XEN_CONSOLE_TEMPORARY_BUF_SIZE`

xen-console-srv/src/xen_console.c
lorc: Because this is an inter-domain ring buffer. There are no lock by definition. Correctness of operation is guaranteed by correct data barrier use. Please see the `xen_console: use correct barriers` patch for more information.

xen-console-srv/src/xen_console.c
lorc: Do you have particular reason to suggest using locking or atomic ops in this case?

xakep-amatop: race condition between init and attach

xen-console-srv/src/xen_console.c
```diff
@@ -0,0 +1,456 @@
...
+	}
+
+	return 0;
+
+err_unbind:
+	unbind_event_channel(console->local_evtchn);
+	evtchn_close(console->local_evtchn);
+
+err_free:
+	k_free(console->int_buf);
```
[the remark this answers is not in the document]
lorc: If console initialization fails, it is expected that domain instance will be destroyed immediately. So there is no great need to re-assign the pointer.

xen-console-srv/src/xen_console.c
lorc: It is expected that attach will never be called if init is incomplete:

1. This is common pattern: user must ensure that they don't try to use not-initialized modules. Actually, it is very hard to prevent use-before-init from inside a module. For example, in this case you can't use `console->lock` because it may be not initialized yet.

2. In this particular case, domain will not be created if console is not properly initialized. So you can't get valid `struct domain *` pointer to pass it to `attach`

xen-console-srv/src/xen_console.c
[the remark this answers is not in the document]
lorc: If initialization of console fails, domain will not be created at all. User will not have valid `struct domain*` pointer to pass it to `attach`. Also, please note, that you **want** `first_attach` to be `true`, otherwise `attach` function will call `k_thread_attach` on invalid pointer.

xen-console-srv/src/xen_console.c
lorc: Actually, after reworking `get_stack_idx` is guaranteed to return a valid value.

xen-console-srv/src/xen_console.c
xakep-amatop: thanks for explanations

xakep-amatop (approved):

xen-console-srv/src/xen_console.c
```diff
@@ -170,9 +193,14 @@ int xen_init_domain_console(struct xen_domain *domain)
 
 	if (rc) {
 		LOG_ERR("Failed to set domain console evtchn param (rc=%d)", rc);
-		return rc;
+		goto err_free;
```
firscity: Old problem: here should be smth like `goto unbind_evtchn` for `console->local_evtchn`, that was taken previously.

xen-console-srv/src/xen_console.c
```diff
@@ -271,3 +334,118 @@ int xen_stop_domain_console(struct xen_domain *domain)
 
 	return 0;
 }
+
+#ifdef CONFIG_XEN_SHELL
+
+static void console_display_thrd(void *p1, void *p2, void *p3)
+{
+	ARG_UNUSED(p2);
```
firscity, line 344: This line is invalid, if you are using `p2` as shell pointer.

firscity (approved):
Amazing job, thank you!

PR #23, opened 2023-03-29
xenstore-srv: add initcall for xenstore root node initialization

firscity:
Previously, Xenstore server required initialization of root node inside target application main(). It was hack and can be easily fixed with Zephyr initcall inside xenstore-srv.

Now root initialization is performed on APPLICATION Zephyr init level, so applications now have no need to care about xenstore-srv.


LKomaryanskiy (approved):

xakep-amatop (approved):

PR #24, opened 2023-03-29
!fixup domain.h: Add callback interface for image loading

firscity:
This commit fixes issue in probe_zimage, related to incorrect usage of offset parameter. It should be used only for pointing to src address of image, that is about to be loaded (not to calculate load addr etc.). Previous version broke domain bootup when received uhdr offset as this param and added it to load address.

Also added proper naming and function callback description.


xakep-amatop (approved):

LKomaryanskiy (approved):
Good job!

PR #26, opened 2023-03-31
xen_console: fix crash while destroying a domain

lorc:
There is a possible issue with attached console when domain is being destroyed in background. Basically, xen_stop_domain_console() sets current_console to NULL and then console_shell_cb() crashes on assertion.

Correct behavior for console_shell_cb() is to disable callback and exit in a such case.


xakep-amatop (approved):

LKomaryanskiy:

firscity (approved):

PR #30, opened 2023-04-10
Refactor Xenstore events

firscity:


xenstore-srv/src/xenstore_srv.c
```diff
@@ -293,42 +293,55 @@ static void send_errno(struct xen_domain *domain, uint32_t id, int err)
...
-				klen = loclen;
-			else {
-				klen = 0;
-				ioffset = 0;
-				ooffset = 1;
-			}
+	pendkey_len = strlen(pending_path);
+
+	snprintf(local, sizeof(local), "/local/domain/%d/", domain->domid);
+	loc_len = strlen(local);
```
xakep-amatop, line 306: (note) this call is redundant just use return value from the prev call

xenstore-srv/src/xenstore_srv.c
```diff
@@ -293,42 +293,55 @@ static void send_errno(struct xen_domain *domain, uint32_t id, int err)
...
 
-			size_t tlen = strlen(iter->token);
-			size_t plen = tlen + kplen - klen + 1 + ooffset;
-			char *pload = k_malloc(plen);
+	/* This function should be called when we already hold wel_mutex */
+	SYS_DLIST_FOR_EACH_CONTAINER_SAFE (&watch_entry_list, iter,
```
xakep-amatop, line 309: (question) why have you used `_SAFE` iterator?
as far as I see you haven't deleted items from list

xenstore-srv/src/xenstore_srv.c
firscity: Thanks, done

xenstore-srv/src/xenstore_srv.c
firscity: Yeah, you're right, it is redundant. Fixed.

LKomaryanskiy:
Good job!

xenstore-srv/src/xenstore_srv.c
```diff
@@ -293,42 +296,53 @@ static void send_errno(struct xen_domain *domain, uint32_t id, int err)
...
-				klen = 0;
-				ioffset = 0;
-				ooffset = 1;
-			}
+	pendkey_len = strlen(pending_path);
 
-			size_t tlen = strlen(iter->token);
-			size_t plen = tlen + kplen - klen + 1 + ooffset;
-			char *pload = k_malloc(plen);
+	loc_len = snprintf(local, sizeof(local), "/local/domain/%d/",
```
lorc, line 307: Need to check `lock_len`, because `snprintf` can return value greater than `sizeof(local)`.  For some unknown reason our `domain->domid` has type of `uint32_t`, so potentially you can get very wrong result there.

lorc (commented):
s/sequance/sequence in the commit message

xenstore-srv/src/xenstore_srv.c
firscity: Thanks, added an assertion here. But probably, domid type should be also fixed in further PR.

lorc (commented):

PR #31, opened 2023-04-11
XRUN update parameters due to the integration results

oleksiimoisieiev:
This MR includes 2 changes:
1) Use correct path size for kernel and dt images, received in runtime spec;
2) Use the same GIC version as on the system. It appears that Spider board do not support compatibility between GIC versions and sends not supported error when trying to start domain with gicv2 on it. That's why using system default.


xakep-amatop (approved):

firscity (approved):
s/back compatibility/backward compatibility

PR #32, opened 2023-04-13
xen_console_srv: make xen console server optional

lorc:
PV console is actually an optional feature. It can be disabled if user is completely sure that DomUs will never use it, or if all DomUs can handle console ring buffer overrun. So, it is strictly advised to enable it, but we need to give user an ability to disable it.

Thus, make all xen_console dependencies optional, remove dependency on XEN_CONSOLE_SRV from XEN_DOMAIN_MANAGEMENT and remove "default y".

Suggested-by: Oleksandr Grytsov <oleksandr_grytsov@epam.com>

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -26,14 +26,12 @@ config PARTIAL_DEVICE_TREE_SIZE
 
 config XEN_DOMAIN_MANAGEMENT
 	bool "Enable Xen Domain Management"
-	depends on XEN_CONSOLE_SRV
 	help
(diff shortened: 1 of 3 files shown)


lorc:
@al1img, could you please confirm that this patch provides functionality you requested?

al1img:
`Reviewed-by: Oleksandr Grytsov <oleksandr_grytsov@epam.com>`
`Tested-by: Oleksandr Grytsov <oleksandr_grytsov@epam.com>`

PR #43, opened 2023-05-25
mem-mgmt: update xen-mgmt to use xen extended regions

oleksiimoisieiev:
Using xen extended regions if enabled to get buffer for mapping domain pages during domain creation. When the Domain is created, it's memory should be temporary mapped to Dom0 to put some data, such as kernel, etc. Using xen extended regions driver allows us to avoid allocating memory for mapping in heap.

### DIFF
--- a/xen-dom-mgmt/src/mem-mgmt.c
+++ b/xen-dom-mgmt/src/mem-mgmt.c
@@ -5,6 +5,9 @@
 
 #include <zephyr/xen/dom0/domctl.h>
 #include <zephyr/xen/memory.h>
+#if defined(CONFIG_XEN_REGIONS)
+#include <zephyr/xen/regions.h>
+#endif
 #include <zephyr/logging/log.h>
 
 #include <mem-mgmt.h>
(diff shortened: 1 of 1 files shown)


xen-dom-mgmt/src/mem-mgmt.c
```diff
@@ -195,6 +184,49 @@ int xenmem_unmap_region(uint64_t nr_pages, void *mapped_addr)
...
+							      nr_pages);
+
+	if (nr_added_pfns != nr_pages) {
+		goto err_out;
+	}
+
+	return 0;
+
+err_out:
+	ret = put_region_space(*mapped_addr, nr_pages);
```
firscity, line 218: Passing `nr_pages` instead of `nr_added_pfns` will break old recover scenario. Please, check it out.

firscity (commented):
In commit message `it's` -> `its`.

xen-dom-mgmt/src/mem-mgmt.c
xakep-amatop, line 218: (cosmetic note) I see no reason to have `ret` variable you can just call this function inside `if`

xen-dom-mgmt/src/mem-mgmt.c
```diff
@@ -195,6 +184,49 @@ int xenmem_unmap_region(uint64_t nr_pages, void *mapped_addr)
...
+	uint64_t nr_added_pfns, base_pfn;
+	int rc = -ENOMEM, ret;
+
+	if (!mapped_addr) {
+		return -EINVAL;
+	}
+
+	*mapped_addr = get_region_space(nr_pages);
+	if (!*mapped_addr) {
+		LOG_ERR("Failed to alloc memory for mapping");
```
xakep-amatop, line 201: (note) I think nr_pages in logs will be useful

xen-dom-mgmt/src/mem-mgmt.c
```diff
@@ -195,6 +184,49 @@ int xenmem_unmap_region(uint64_t nr_pages, void *mapped_addr)
...
+	}
+
+	base_pfn = xen_virt_to_gfn(*mapped_addr);
+	nr_added_pfns = xendom_add_to_physmap_batch_by_chunks(domid,
+							      base_pfn,
+							      base_gfn,
+							      nr_pages);
+
+	if (nr_added_pfns != nr_pages) {
+		goto err_out;
```
xakep-amatop, line 212: (question) should you add log message here?

xakep-amatop, line 212: (question) should you remove from physmap currently added pages (difference)?

xen-dom-mgmt/src/mem-mgmt.c
```diff
@@ -195,6 +184,49 @@ int xenmem_unmap_region(uint64_t nr_pages, void *mapped_addr)
...
+	}
+
+	*mapped_addr = get_region_space(nr_pages);
+	if (!*mapped_addr) {
+		LOG_ERR("Failed to alloc memory for mapping");
+		return -ENOMEM;
+	}
+
+	base_pfn = xen_virt_to_gfn(*mapped_addr);
+	nr_added_pfns = xendom_add_to_physmap_batch_by_chunks(domid,
```
xakep-amatop, line 206: (note) you can make one line here it will be exactly 100 symbols

xen-dom-mgmt/src/mem-mgmt.c
xakep-amatop, line 212: (question) Is ENOMEM status here ok?

xen-dom-mgmt/src/mem-mgmt.c
oleksiimoisieiev: Fixed by splitting put_region for 2 calls. Please take a look, maybe you can think of better approach

xen-dom-mgmt/src/mem-mgmt.c
xakep-amatop: maybe the reason why we can't allocate is a huge amount of pages, I think it will be useful hint

xen-dom-mgmt/src/mem-mgmt.c
oleksiimoisieiev: @LKomaryanskiy  @xakep-amatop Guys please take a look on my refactoring. It will fix an issue, but I'm not sure in the approach itselft. Maybe you have some thoughts how it can be made better?

oleksiimoisieiev: @LKomaryanskiy I've posted recently an updated commit, that does pretty much the same. I've split put_pages to 2 calls.

LKomaryanskiy: Yep, I checked the latest version of PR and my comment refers to the latest changes. Maybe not for exactly these lines, but about the main logic of memory-restoring code. It's better to look [here](https://github.com/xen-troops/zephyr-xenlib/pull/43/commits/d04e0c32290fb17ee3cdfd900239dc5472b49fc5#diff-7b2d59a4d9ef29d5b809c8eb280de80f97499e430a1031fa6c3d33e25668dff8R237) - my previous comment about this line.

LKomaryanskiy (approved):
@oleksiimoisieiev good job!
`Reviewed-by: Leonid Komarianskyi <leonid_komarianskyi@epam.com>`

firscity (approved):

oleksiimoisieiev:
Updated

oleksiimoisieiev:
@firscity @oleksiimoisieiev @xakep-amatop Added small fix to prevent build warning when CONFIG_XEN_REGIONS=y

PR #44, opened 2023-06-01
vch: guard vchannel handle from uninitialized access

sa-kib:
Zero client/server vchannel handles on error, so that no dangling pointers are left in the structure.

This is done in case vch_read()/vch_write() is accidentally called on not opened vchannel, and some non-zero non-valid leftover pointers (like h->ring) confuse validation logic.

Also validate the ring pointer when checking for opened channel.


oleksiimoisieiev:
+1 Acked-by: Oleksii Moisieiev <oleksii_moisieiev@epam.com>

vch/src/vch.c
```diff
@@ -236,6 +240,7 @@ int vch_open(domid_t domain, const char *path, size_t min_rs, size_t min_ws,
 	if (h->ring) {
 		k_free(h->ring);
 	}
+	memset(h, 0, sizeof(*h));
```
xakep-amatop, line 243: (question) should you free ring buffer and cleanup `h` in the next case:
```

	rc = alloc_unbound_event_channel(domain);
	if (rc < 0) {
		return rc;
	}
```
?

sa-kib: made a patch with fix [9402fbb](https://github.com/xen-troops/zephyr-xenlib/pull/44/commits/9402fbb934de47f0cc1ba67d89d6ab8bff64c15e), thank you

xakep-amatop (approved):

sa-kib:
rebased + tags added

oleksiimoisieiev:
@sa-kib I see chechpatch errors in this MR

PR #51, opened 2023-12-08
domain: dom-fdt: add machine-specific board compatible to domain config

firscity:
Some device drivers require correct root node compatible during initialization process (e.g. get board revision). Previously all domain device trees contained only "xen,xenvm" compatible which is not correct.

Add possibility to configure this field via domain configuration using "machine_dt_compat" field. It will be used during domain device-tree creation.

### DIFF
--- a/xen-dom-mgmt/src/xen-dom-fdt.c
+++ b/xen-dom-mgmt/src/xen-dom-fdt.c
@@ -113,7 +113,7 @@ static int fdt_property_interrupts(void *fdt, gic_interrupt *intr,
 	return 0;
 }
 
-static int create_root(int major, int minor, void *fdt)
+static int create_root(int major, int minor, void *fdt, struct xen_domain_cfg *domcfg)
 {
 	int res;
 	char buf[FDT_STRING_MAX];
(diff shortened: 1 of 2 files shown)


LKomaryanskiy:
Good job!

xen-dom-mgmt/src/xen-dom-fdt.c
```diff
@@ -130,7 +130,13 @@ static int create_root(int major, int minor, void *fdt)
 	if (!res)
 		return -ENOMEM;
 
-	res = fdt_property_compat(fdt, 2, buf, "xen,xenvm");
+	/* Check if custom machine compatible is not empty and use it */
+	if (domcfg->machine_dt_compat) {
+		res = fdt_property_compat(fdt, 2, buf, domcfg->machine_dt_compat);
```
lorc, line 135: Just a note.

This is not ideal, because generally we can have multiple compat strings:

`"renesas,spider-breakout","renesas,spider-cpu","renesas,r8a779f0"`

but for now it is okay, as this is simpler API.

lorc (approved):

arminn:
Acked-by: Andrii Chepurnyi <andrii_chepurnyi@epam.com>

PR #55, opened 2024-03-18
xen_cmds: use DomU config in case of creating user domain

xakep-amatop:
Use DomU config instead of DomD for creating domains with domid higher 1.

### Related PR: https://github.com/aoscloud/aos_core_zephyr/pull/138


xakep-amatop:
we don't need to support DomUs creation in AOS repo, so I'm closing this PR

PR #57, opened 2024-04-01
xen-shell-cmd: Allow config selection for domain creation

Deedone:
Add linker script to collect all domain configs in one place. Add macro DECL_CONFIG to mark domain config structures. Specific config can be selected for domain creation by name via the shell command line parameter.


xen-shell-cmd/src/xen_cmds.c
```diff
@@ -137,8 +174,8 @@ SHELL_STATIC_SUBCMD_SET_CREATE(
 	subcmd_xu,
 	SHELL_CMD_ARG(create, NULL,
 		      " Create Xen domain\n"
-		      " Usage: create -d <domid>\n",
-		      domu_create, 3, 0),
+		      " Usage: create -d <domid> -c <config_name>\n",
```
firscity, line 177: IMHO, if we make this option as mandatory, then we need to have some kind of `xu config_list` command

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -32,29 +34,64 @@ uint32_t parse_domid(size_t argc, char **argv)
...
-	 */
-	ret = domain_create(&domd_cfg, domid);
+
+	name = parse_name(argc, argv);
+	if (!name) {
+		shell_error(shell, "Invalid config passed to create cmd");
+		return -EINVAL;
+	}
+
+	for (cfg = _domain_configs_start; cfg < _domain_configs_end; cfg++) {
```
firscity, line 74: I think it would be better to move section creating/linking/parsing to xen-dom-mgmt and call smth like
`domain_cfg *dom = xen_dom_get_config(name);` + make it as a Kconfig

It will help to use it from other part of zephyr-xenlib or applications, domains can be created directly via `domain_create(dom_cfg)` as it made for AoS

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -15,7 +16,8 @@
 
 LOG_MODULE_REGISTER(xen_shell);
 
-extern struct xen_domain_cfg domd_cfg;
```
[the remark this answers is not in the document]
Deedone: The intention of this change is to remove the direct dependency on external variable. Section attributes should be added in the same place where the config struct is declared, which is currently in the AOS project.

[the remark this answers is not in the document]
xakep-amatop: I understand. I mean, could you open a pull request with the necessary changes so that we can use the 'xu create' command with `domD` as well?

xen-shell-cmd/src/xen_cmds.c
lorc: Do we really need `-c` part at all? This is mandatory option anyways

oleksiimoisieiev:
Checkpatch check was failed

Deedone:
> Checkpatch check was failed

It is a false positive.

GrygiriiS:
Why not create _init(struct xen_domain_cfg*) for library? It really should be problem of Zephyr app and not a library.

Deedone:
Depends on https://github.com/xen-troops/zephyr-xenlib/pull/74

Deedone:
> Why not create _init(struct xen_domain_cfg*) for library? It really should be problem of Zephyr app and not a library where to place/get dom_cfgs.

This will force every application developer to come up with a similar system. I believe it is better to hide the complexity within a library, instead of forcing it onto its users.

GrygiriiS:
> > Why not create _init(struct xen_domain_cfg*) for library? It really should be problem of Zephyr app and not a library where to place/get dom_cfgs.
> 
> This will force every application developer to come up with a similar system. I believe it is better to hide the complexity within a library, instead of forcing it onto its users.

It forces applications to use some linker sections which in turn  adds:
- platform specific issues/dependencies like cons & not const, XIP vs not XIP
- prevents placing/retrieving  dom_cfg form anywhere like shared mem, or placing cfg in FIP image for example (FIP can be attached to Zephyr binary and prepared outside of Zephyr build system)
- prevents form building and using dynamic cfg list (i do not like it but could happen)

Playing with linker in Zephyr is powerful tool from one side, but painful from another - ask @oleksiimoisieiev.

Really, simplest possible solution for library is xenlib_int().

lorc:
I agree with @GrygiriiS here. App should provide own callbacks for obtaining configs, images, etc. We can provide some generic implementation in xenlib for a default use case. But it should be strictly optional.

GrygiriiS:
> I agree with @GrygiriiS here. App should provide own callbacks for obtaining configs, images, etc. We can provide some generic implementation in xenlib for a default use case. But it should be strictly optional.

Hope I understood you idea right, add
```
typedef int  (*domain_get_cfg_cb_t)(const uint32_t domid,  const struct xen_domain_cfg **dom_cfg);

int domain_set_get_cfg_cb(domain_get_cfg_cb_t cb);
```
The main user for now is XEN shell. A small question about func naming, but I like the Idea.

Deedone:
Added additional way for apps to provide their configs
Added config option to disable additional section creation

Waiting for merge of https://github.com/xen-troops/zephyr-xenlib/pull/74 to update cmdline interface

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -151,6 +198,9 @@ SHELL_STATIC_SUBCMD_SET_CREATE(
 		      " Unpause Xen domain\n"
 		      " Usage: unpause -d <domid>\n",
 		      domu_unpause, 3, 0),
+	SHELL_CMD_ARG(config_list, NULL,
+		      " List available domain configurations\n",
+		      xen_config_list, 1, 0),
```
GrygiriiS, line 203: How about "config list"? With possibility to have "config dump" in the future, for example?

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -133,12 +158,34 @@ int domu_unpause(const struct shell *shell, size_t argc, char **argv)
...
+{
+	__maybe_unused struct xen_domain_cfg *cfg;
+	int i;
+
+	ARG_UNUSED(argc);
+	ARG_UNUSED(argv);
+
+	for (i = 0; i < domain_get_user_cfg_count(); i++) {
+		cfg = domain_get_user_cfg(i);
+		shell_print(shell, "%s", cfg->name);
```
GrygiriiS, line 171: Could you add index in print, pls? "%d %s" <- i, name

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -133,12 +158,34 @@ int domu_unpause(const struct shell *shell, size_t argc, char **argv)
...
+	}
+
+#ifdef CONFIG_XEN_DOMCFG_SECTION
+	for (cfg = _domain_configs_start; cfg < _domain_configs_end; cfg++) {
+		shell_print(shell, "%s", cfg->name);
```
GrygiriiS, line 176: Could you add index in print, pls? "%d %s" <- i, name

xen-shell-cmd/src/xen_cmds.c
Deedone: Thank you for the suggestion but I am not sure if this change will provide any actual benefit.

xen-shell-cmd/src/xen_cmds.c
Deedone: No, because it will add confusion suggesting to users that they can somehow use this number somewhere.

xen-shell-cmd/src/xen_cmds.c
Deedone: Same

GrygiriiS:
`Tested-by: Grygorii Strashko <grygorii_strashko@epam.com>`

xen-dom-mgmt/linker.ld
```diff
@@ -0,0 +1,6 @@
+.domain_configs :
+{
+  _domain_configs_start = .;
+  KEEP(*(.domain_configs))
+  _domain_configs_end = .;
+} > RAM
```
lorc, line 6: In which section it will reside at the end? I expect it to get somewhere in `.rodata`

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -616,6 +617,21 @@ struct xen_domain *domid_to_domain(uint32_t domid)
 	return NULL;
 }
 
+struct xen_domain_cfg *domain_find_config(char *name)
```
lorc, line 620: `const char *name`

xen-dom-mgmt/linker.ld
GrygiriiS: no :P Not exactly :) Now it can't be made const. More over, it depends on iomems, dtdevs, dt_passthrough.
Now it can be only in RAM

I kinda did experiment to build dom_cfg.c with one config as .o - results are not encouraging.

lorc: It's a pity. On other hand, I can imagine that someone might want to alter configuration in run-time. So maybe, in the end, this is not as bad as it seems.

firscity: Also, `back_cfg` is stored in `xen_domain_cfg` struct, it is filled manually by app or via xrun external configuration before domain is created. Thus we definitely need it to be r/w or rework xrun + app processing to separate structure.

Deedone: It is at the end of data_region

lorc (approved):
Thanks

firscity (approved):

PR #63, opened 2024-04-09
xen-dom-mgmt: dom0: add dom0 initcall routine

firscity:
During Zephyr xenlib initialization some actions (root node init in Xenstore, Domain-0 Xenstore entries adding) should be done right before application start. Previously these actions were implemented in separate initcalls with different priorities which is a little bit complicated.

As we will need to have Domain-0 struct in runtime soon (for further implementation of xss watch functions in Xenstore API), Domain-0 related actions were moved to single routine: root node init was made global and xenstore entries adding was refatored. So now we can simplify priorities for initcalls and have a single place where all pre-application actions can be done.

### DIFF
--- a/xen-dom-mgmt/src/xen-dom-mgmt.c
+++ b/xen-dom-mgmt/src/xen-dom-mgmt.c
@@ -899,3 +899,64 @@ deinit:
 	domain_destroy(domid);
 	return rc;
 }
+
+static int init_domain0(const struct device *d)
+{
+	ARG_UNUSED(d);
+	int ret = 0;
+	struct xen_domain *dom0 = NULL;
+#ifdef CONFIG_XSTAT
+	struct xenstat_domain *dom0stat;
+
+	dom0stat = k_malloc(sizeof(*dom0stat));
+	if (!dom0stat) {
+		ret = -ENOMEM;
+		LOG_ERR("Can't allocate memory for dom0 xenstat");
+		goto out;
+	}
+
+	ret = xstat_getdominfo(dom0stat, 0, 1);
+	if (ret < 0) {
+		LOG_ERR("Failed to get info for dom0 (rc=%d)", ret);
+		goto out;
+	} else if (ret == 0) {
+		/* Theoretically impossible */
+		ret = -EINVAL;
+		goto out;
+	}
+#endif
+
+	dom0 = k_malloc(sizeof(*dom0));
+	if (!dom0) {
+		ret = -ENOMEM;
+		LOG_ERR("Can't allocate memory for dom0 domain struct");
+		goto out;
+	}
+	memset(dom0, 0, sizeof(*dom0));
+
+	snprintf(dom0->name, CONTAINER_NAME_SIZE, "%s", DOM0_NAME);
+	dom0->domid = 0;
+
+#ifdef CONFIG_XSTAT
+	dom0->num_vcpus = dom0stat->num_vcpus;
+	dom0->max_mem_kb = dom0stat->cur_mem / 1024;
+	k_free(dom0stat);
+#else
+	dom0->num_vcpus = 0;
+	dom0->max_mem_kb = 0;
+#endif
+
+	xs_init_root();
+	xss_write("/tool/xenstored", "");
+	xs_initialize_xenstore(0, dom0);
+
+out:
+#ifdef CONFIG_XSTAT
+	k_free(dom0stat);
+#endif
+	k_free(dom0);
+
+	return ret;
+}
+
+SYS_INIT(init_domain0, APPLICATION, CONFIG_KERNEL_INIT_PRIORITY_DEFAULT);
--- a/xen-dom-mgmt/src/xen-dom-xs.c
+++ b/xen-dom-mgmt/src/xen-dom-xs.c
@@ -16,7 +16,6 @@
 #include <domain.h>
 #include <xen/public/io/xs_wire.h>
 #include <xss.h>
-#include <xstat.h>
 #include <xen-dom-xs.h>
 #include <xenstore_srv.h>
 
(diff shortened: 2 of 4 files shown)


lorc (approved):

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -899,3 +899,64 @@ int domain_post_create(const struct xen_domain_cfg *domcfg, uint32_t domid)
...
 	return rc;
 }
+
+static int init_domain0(const struct device *d)
+{
+	ARG_UNUSED(d);
+	int ret = 0;
+	struct xen_domain *dom0 = NULL;
+#ifdef CONFIG_XSTAT
+	struct xenstat_domain *dom0stat = NULL;
```
dsemenets, line 909: Initialization unnecessarily here

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -899,3 +899,64 @@ int domain_post_create(const struct xen_domain_cfg *domcfg, uint32_t domid)
 	domain_destroy(domid);
 	return rc;
 }
+
+static int init_domain0(const struct device *d)
+{
+	ARG_UNUSED(d);
+	int ret = 0;
+	struct xen_domain *dom0 = NULL;
```
dsemenets, line 907: Initialization unnecessarily here

firscity: I left this part as it was. Actually this exact initialization is not really needed, because it will be initialized in next lines.

xen-dom-mgmt/src/xen-dom-mgmt.c
firscity: I left this part as it was, this initialization is needed. In case of enabled `CONFIG_XENSTAT` and error in `k_malloc` or `xstat_getdominfo` we will pass stack garbage value in k_free().

dsemenets:
`Acked-by: Dmytro Semenets <dmytro_semenets@epam.com>`

xen-dom-mgmt/src/xen-dom-mgmt.c
firscity: Removed in new version

firscity:
Remove unnecessary `*dom0stat = NULL` initialization
Add @lorc and @dsemenets Acked-by's

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -899,3 +899,64 @@ int domain_post_create(const struct xen_domain_cfg *domcfg, uint32_t domid)
...
+	dom0->max_mem_kb = dom0stat->cur_mem / 1024;
+	k_free(dom0stat);
+#else
+	dom0->num_vcpus = 0;
+	dom0->max_mem_kb = 0;
+#endif
+
+	xs_init_root();
+	xss_write("/tool/xenstored", "");
+	xs_initialize_xenstore(0, dom0);
```
xakep-amatop, line 951: In the previous version of the code, we stored the return result in the ret variable and returned it from the current function. However, in the current implementation, we return some positive non-zero value. I'm afraid that it could impact the initialization procedure under SYS_INIT in Zephyr. I've investigated the initialization logic and see no problem here because the ret result is used only in cases when we have a valid device pointer. However, we don't know what will be in future releases of Zephyr; at least, this behavior is undocumented.

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -899,3 +899,64 @@ int domain_post_create(const struct xen_domain_cfg *domcfg, uint32_t domid)
...
+	dom0->num_vcpus = dom0stat->num_vcpus;
+	dom0->max_mem_kb = dom0stat->cur_mem / 1024;
+	k_free(dom0stat);
+#else
+	dom0->num_vcpus = 0;
+	dom0->max_mem_kb = 0;
+#endif
+
+	xs_init_root();
+	xss_write("/tool/xenstored", "");
```
xakep-amatop, line 950: should we check result before calling `xs_initialize_xenstore`?

xen-dom-mgmt/src/xen-dom-mgmt.c
firscity: thanks, will change it to just 0

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -899,3 +899,64 @@ int domain_post_create(const struct xen_domain_cfg *domcfg, uint32_t domid)
...
+	xs_initialize_xenstore(0, dom0);
+
+out:
+#ifdef CONFIG_XSTAT
+	k_free(dom0stat);
```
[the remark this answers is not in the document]
firscity: yeah, looks like part of further my change, will fix it in separate PR

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -899,3 +899,64 @@ int domain_post_create(const struct xen_domain_cfg *domcfg, uint32_t domid)
...
+#endif
+	k_free(dom0);
+
+	return ret;
+}
+
+SYS_INIT(init_domain0, APPLICATION, CONFIG_KERNEL_INIT_PRIORITY_DEFAULT);
```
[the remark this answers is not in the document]
firscity: It is not actually matter since it only need to be called somewhere before application `main`

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -899,3 +899,64 @@ int domain_post_create(const struct xen_domain_cfg *domcfg, uint32_t domid)
...
+
+	snprintf(dom0->name, CONTAINER_NAME_SIZE, "%s", DOM0_NAME);
+	dom0->domid = 0;
+
+#ifdef CONFIG_XSTAT
+	dom0->num_vcpus = dom0stat->num_vcpus;
+	dom0->max_mem_kb = dom0stat->cur_mem / 1024;
+	k_free(dom0stat);
+#else
+	dom0->num_vcpus = 0;
```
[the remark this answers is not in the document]
xakep-amatop: so, can we delete memset?

xen-dom-mgmt/src/xen-dom-mgmt.c
xakep-amatop: P.S. I'm not sure if MISRA rules are applicable to xenlib or not

PR #66, opened 2024-04-12
xen-console: Optimize console_display_thrd

Deedone:
Optimize console_display_thrd by removing the inner loop that copies data from the console buffer to the output buffer. Instead, calculate the size of the data to be copied and print it in one go.

### DIFF
--- a/xen-console-srv/src/xen_console.c
+++ b/xen-console-srv/src/xen_console.c
@@ -367,9 +367,8 @@ static void console_display_thrd(void *p1, void *p2, void *p3)
 	ARG_UNUSED(p3);
 	struct xen_domain_console *console = p1;
 	const struct shell *shell = p2;
-	/* Buffer input a little */
-	char buf[32];
-	int read;
+	size_t buf_pos, prod_buf_pos;
+	int size;
 
 	shell_info(shell, "Attached to a domain console");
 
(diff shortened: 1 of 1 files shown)


xen-console-srv/src/xen_console.c
```diff
@@ -387,19 +386,13 @@ static void console_display_thrd(void *p1, void *p2, void *p3)
...
-			while (console->int_cons < console->int_prod &&
-			       read < sizeof(buf) - 1) {
-				size_t buf_pos = (console->int_cons++) &
-					(XEN_CONSOLE_BUFFER_SZ - 1);
-				buf[read++] = console->int_buf[buf_pos];
-			}
-			if (read) {
-				shell_fprintf(shell, SHELL_NORMAL, "%s", buf);
-			}
+			buf_pos = (console->int_cons++) &
```
firscity, line 389: Increment should be removed

xen-console-srv/src/xen_console.c
```diff
@@ -387,19 +386,13 @@ static void console_display_thrd(void *p1, void *p2, void *p3)
...
+				(XEN_CONSOLE_BUFFER_SZ - 1);
+			size = (console->int_prod - console->int_cons) &
+				(XEN_CONSOLE_BUFFER_SZ - 1);
+			shell_fprintf(shell, SHELL_NORMAL, "%.*s", size + 1,
+						&console->int_buf[buf_pos]);
```
firscity, line 394: `fprintf` may overrun the end of `int_buf` if `buf_pos` is closer than `size+1` bytes to end of array. This will cause access to foreign memory, so this should be normalized and divided into 2 parts in this corner case.

lorc: Yep. This is a circular buffer so wrap around should be handled gracefully.

Deedone:
Fixed logic to prevent buffer overruns.

xen-console-srv/src/xen_console.c
```diff
@@ -387,19 +386,18 @@ static void console_display_thrd(void *p1, void *p2, void *p3)
...
-					(XEN_CONSOLE_BUFFER_SZ - 1);
-				buf[read++] = console->int_buf[buf_pos];
-			}
-			if (read) {
-				shell_fprintf(shell, SHELL_NORMAL, "%s", buf);
+			buf_pos = (console->int_cons) &
+				(XEN_CONSOLE_BUFFER_SZ - 1);
+			prod_buf_pos = (console->int_prod) &
+				(XEN_CONSOLE_BUFFER_SZ - 1);
+			if (buf_pos < prod_buf_pos) {
```
firscity, line 393: Please, comment that this condition means no buffer wrap around and in other case we print only part before buffer end

firscity (commented):
With last small comment:

lorc (approved):
With @firscity comment addressed:

firscity (approved):

PR #67, opened 2024-04-16
xrun: Fix potential concurrency issues

Deedone:
Add reference counting to container objects to prevent unexpected failures when container object is freed right after return from get_container function.
Add locking to pause/resume functions.

### DIFF
--- a/xrun/src/xrun.c
+++ b/xrun/src/xrun.c
@@ -96,6 +96,7 @@ struct container {
 	char dt_image[CONFIG_XRUN_MAX_PATH_SIZE];
 	bool has_dt_image;
 	enum container_status status;
+	int refcount;
 };
 
 static const struct json_obj_descr hypervisor_spec_descr[] = {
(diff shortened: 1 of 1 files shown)


xrun/src/xrun.c
```diff
@@ -217,23 +230,25 @@ static struct container *register_container_id(const char *container_id)
...
 	return container;
 }
 
 static int unregister_container_id(const char *container_id)
 {
-	struct container *container = get_container(container_id);
+	struct container *container = NULL;
+
+	k_mutex_lock(&container_lock, K_FOREVER);
+	container = find_container_locked(container_id);
```
lorc, line 244: This is prone to ABA problem. Reference owner should hold own pointer to the container.

xrun/src/xrun.c
```diff
@@ -635,10 +650,12 @@ int xrun_pause(const char *container_id)
 
 	ret = domain_pause(container->domid);
```
lorc, line 651: Here is another issue, which is related: all this pause/unpause stuff should be done while holding container own lock.

We need another patch to fix this.

xrun/src/xrun.c
```diff
@@ -671,9 +690,11 @@ int xrun_kill(const char *container_id)
 
 	ret = domain_destroy(container->domid);
 	if (ret) {
+		put_container(container);
 		return ret;
 	}
 
+	put_container(container);
 	return unregister_container_id(container_id);
```
lorc, line 698: You can just call `put_container` for a second time here. With a comment, of course.

xrun/src/xrun.c
oleksiimoisieiev: I support domain_pause/unpause is responsible for this lock. Because to be honest this lock should be domid own lock rather then container related.

xrun/src/xrun.c
```diff
@@ -653,10 +670,12 @@ int xrun_resume(const char *container_id)
...
 	ret = domain_unpause(container->domid);
 	if (ret) {
+		put_container(container);
 		return ret;
 	}
 
 	container->status = RUNNING;
+	put_container(container);
 	return 0;
 }
```
oleksiimoisieiev, line 680:
```suggestion
	ret = domain_unpause(container->domid);
	if (ret) {
		goto out;
	}

	container->status = RUNNING;
out:
	put_container(container);
	return ret;
}
```

oleksiimoisieiev: Not mandatory. Just will look more readable from my standpoint. The same suggestion across all put_container calls

oleksiimoisieiev (changes requested):

xrun/src/xrun.c
Deedone: Is it not ok? We can be sure that the container pointer is valid, and if two different threads try to write to `container->status` at the same time, one of those writes would be overwritten by the other in any case.

lorc:
```
    Thread-1   |   Thread-2
               |
domain_pause() |
               |
           CTX SWITCH
               |
               | domain_unpause()
               | state = unpaused
               |
           CTX SWITCH
               |
state = paused |
               |
```

GrygiriiS:
Similar comment as https://github.com/xen-troops/zephyr-xenlib/pull/72#issuecomment-2069257869

xrun/src/xrun.c
```diff
@@ -175,34 +176,37 @@ static struct container *get_container(const char *container_id)
...
-	 * This should be applied to all get_container
-	 * calls.
-	 *
-	 * There are two ways to handle this situation:
-	 * - Hold the global lock the whole time you are
-	 * holding pointer to a container.
-	 * - Add reference counting (like linux kref) to
-	 * ensure that this object does not disappear under
-	 * your feet.
-	 */
```
lorc, line 203: It is a pity, but this function has another concurrency issue: code below checks if container exists by using `get_container`. Problem is that if `register_container_id` is called twice with the same container id, it may register both containers. You need to hold the container lock whole time: obtain the lock, check if container already exists, create new one, free the lock.

This is another issue, so you can add TODO for now. Or make a separate patch that fixes this.

xrun/src/xrun.c
```diff
@@ -671,10 +665,14 @@ int xrun_kill(const char *container_id)
 
 	ret = domain_destroy(container->domid);
 	if (ret) {
-		return ret;
+		goto out;
 	}
 
-	return unregister_container_id(container_id);
+	/* Put container twice to drop the last reference */
+	put_container(container);
```
lorc, line 678: Hm... should you call `domain_destroy` when you are freeing container? Or it will be okay to have a container object without alive domain inside?

lorc (commented):
That should to. Thanks.

How you tested this, BTW?

lorc:
Oh, and  don't forget to fix checkpatch.

Deedone:
Tested with unit tests and trying to start/stop/pause/unpause the aos monkey container.

firscity (approved):
Thank you

PR #72, opened 2024-04-19
xen-dom-mgmt: Fix potential concurrency issues

Deedone:
Add refcounting to the domain structure to make it thread-safe. Fix possible deadlock in console deinit.

### DIFF
--- a/xen-dom-mgmt/src/xen-dom-mgmt.c
+++ b/xen-dom-mgmt/src/xen-dom-mgmt.c
@@ -809,7 +809,7 @@ int domain_destroy(uint32_t domid)
 		return -ENOTSUP;
 	}
 
-	rc = xs_remove_xenstore_backends(domid);
+	rc = xs_remove_xenstore_backends(domain);
 	if (rc) {
 		LOG_ERR("Failed to remove_xenstore_backends domain#%u (rc=%d)", domain->domid, rc);
 		err = rc;
--- a/xen-dom-mgmt/src/xen-dom-mgmt.c
+++ b/xen-dom-mgmt/src/xen-dom-mgmt.c
@@ -579,23 +579,64 @@ static int assign_dtdevs(int domid, char *dtdevs[], int nr_dtdevs)
 	return rc;
 }
 
-/*
- * TODO: Access to domain_list and domains should be protected, considering that it may be
- * destroyed after receiving pointer to actual domain. So all accesses to domains structs should be
- * protected globally or via refcounts. This requires code audit in all libs, that are using this
- * function (currently xenstore-srv and xen_shell).
- */
-struct xen_domain *domid_to_domain(uint32_t domid)
+struct xen_domain *get_domain(uint32_t domid)
 {
 	struct xen_domain *iter;
 
+	k_mutex_lock(&dl_mutex, K_FOREVER);
 	SYS_DLIST_FOR_EACH_CONTAINER (&domain_list, iter, node) {
 		if (iter->domid == domid) {
-			return iter;
+			iter->refcount++;
+			break;
 		}
 	}
+	k_mutex_unlock(&dl_mutex);
+	return iter;
+}
 
-	return NULL;
+void put_domain(struct xen_domain *domain)
+{
+	int rc;
+
+	if (!domain) {
+		LOG_ERR("Domain is NULL");
+		return;
+	}
+	__ASSERT(!domain->f_dom0less, "dom0less domain#%u operation not supported", domain->domid);
+
+	k_mutex_lock(&dl_mutex, K_FOREVER);
+	domain->refcount--;
+	if (domain->refcount == 0) {
+		rc = xs_remove_xenstore_backends(domain);
+		if (rc) {
+			LOG_ERR("Failed to remove_xenstore_backends domain#%u (rc=%d)",
+					domain->domid, rc);
+		}
+
+		rc = stop_domain_stored(domain);
+		if (rc) {
+			LOG_ERR("Failed to stop domain#%u store (rc=%d)", domain->domid, rc);
+		}
+
+		xs_deinitialize_domain_xenstore(domain->domid);
+
+	#ifdef CONFIG_XEN_CONSOLE_SRV
+		rc = xen_stop_domain_console(domain);
+		if (rc) {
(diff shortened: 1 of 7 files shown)


xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -597,23 +597,31 @@ static int assign_dtdevs(int domid, char *dtdevs[], int nr_dtdevs)
...
 
+	k_mutex_lock(&dl_mutex, K_FOREVER);
 	SYS_DLIST_FOR_EACH_CONTAINER (&domain_list, iter, node) {
 		if (iter->domid == domid) {
-			return iter;
+			break;
 		}
 	}
+	if (iter) {
+		k_mutex_lock(&iter->lock, K_FOREVER);
```
firscity, line 611: This can be done right in the condition of upper loop, no need for additional block

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -779,13 +789,20 @@ int domain_destroy(uint32_t domid)
...
 		LOG_ERR("Domain with domid#%u is not found", domid);
 		/* Domain with requested domid is not present in list */
 		return -EINVAL;
 	}
 
+	k_mutex_lock(&domain->lock, K_FOREVER);
+	while (domain->refcount > 1) {
+		LOG_WRN("Domain#%u is still in use, waiting for release (%d refs)",
+						domid, domain->refcount);
+		k_condvar_wait(&domain->refcount_cond, &domain->lock, K_FOREVER);
```
firscity, line 803: I'm not sure if this is right place for this check. IIUC we should wait only before removing domain struct from list and freeing its memory, not before trying to remove backends or console.

firscity (commented):
With 1 more comment:
commit message `Change xs_remove_xenstore_backends() signature` - this is a consequence, not the reason. Please add proper explanation to commit topic (`use domain reference instead of domid for backend init` etc.)

GrygiriiS:
Huh. lock on lock on lock + ref count. It looks very over-complicated and over-designed. In my strong opinion, the locking is problem of Zephyr app and not a library. More over, domain management/control  process should be simple and straight forward in Zephyr app which definitely should not require to have ref_count at minimum. And I think It all should just work with on one global lock.

lorc:
>  domain_destroy routine gets a reference
and waits until it's reference is the only one before destroying the
domain. 

A am not sure I got this logic. `domain_destroy` should put down one reference. Real domain destruction should be performed in `put_domain` when reference counter reaches zero. Take a look at linux `kref` as example.

Deedone:
Moved freeing to put_domain and removed the per-domain lock.

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -597,23 +597,33 @@ static int assign_dtdevs(int domid, char *dtdevs[], int nr_dtdevs)
...
-struct xen_domain *domid_to_domain(uint32_t domid)
+struct xen_domain *get_domain(uint32_t domid)
 {
 	struct xen_domain *iter;
 
+	k_mutex_lock(&dl_mutex, K_FOREVER);
 	SYS_DLIST_FOR_EACH_CONTAINER (&domain_list, iter, node) {
 		if (iter->domid == domid) {
-			return iter;
+			break;
```
firscity, line 607: IIUC `iter->refcount++;` can be added right here, no need for new block

Deedone: Done

GrygiriiS:
@Deedone Could you send first patch alone as it's definitely fix "xen_console: Fix deadlock in xen_stop_domain_console()".

lorc:
> Moved freeing to put_domain and removed the per-domain lock.

What I meant, is that **whole** domain_destroy should be triggered by put_domain(). Because current implementation has no sense: someone may be holding domain reference and trying to do something with it, while you are stopping console, destroying xenstore entries and so on. There is no sense in having reference for a domain, when it basically is destroyed already. At least I see no use case for this. Maybe I am wrong.

I believe @GrygiriiS is right and we have no use for reference counting for now. Maybe global `domain_lock` will be sufficient.  On other hand, we had global `pci_lock` in Xen and it made lots and lots headache when we tried implement more complex logic in PCI subsystem... So I am a little hesitant about introducing huge global locks. 

Anyways, @firscity is maintaining this project and he knows it better, so last word is his.

In the meantime, you can have my:

`Reviewed-by: Volodymyr Babchuk <volodymyr_babchuk@epam.com>` for the first two patches. Could you please create a separate PR with at least first one, so we can merge it faster?

firscity:
> There is no sense in having reference for a domain, when it basically is destroyed already.

Yeah, this part should be changed as you suggested before.

> So I am a little hesitant about introducing huge global locks.

We already have at least 3 ways to create and manage domains - directly from application via `zephyr-xenlib`, via `xrun` json configurations and via Zephyr shell command. I think that global locking (altogether with all other locks in console, xenstore etc.) will definitely bring more problems and will be less flexible than refcounting.
I may be wrong, but I think suggestion with decrementing refcount inside `destroy_domain()` and moving its actual logic to `put_domain()` should work fine.

firscity (commented):
With checkpatch fixed:

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -597,23 +597,67 @@ static int assign_dtdevs(int domid, char *dtdevs[], int nr_dtdevs)
...
+	if (!domain) {
+		LOG_ERR("Domain is NULL");
+		return;
+	}
+
+	k_mutex_lock(&dl_mutex, K_FOREVER);
+	domain->refcount--;
+	if (domain->refcount == 0) {
+		if (domain->f_dom0less) {
+			LOG_ERR("dom0less domain#%u operation not supported", domain->domid);
```
lorc, line 628: I am not fully understand what will happen in this case. We will be stuck with `struct xen_domain` that has `refcount==0` forever? 

Also you need to unlock the mutex.

Shall we replace this with `ASSERT(!domain->f_dom0less)` ? @firscity, @GrygiriiS  ?

Deedone:
Added assert, fixed checkpatch

lorc (approved):

PR #77, opened 2024-05-01
xenstore: rework "xs ls" cmd to dump all entries in a directory recursively 

GrygiriiS:
This series adds new Xenstore xss_list_traverse(0 API which accepts Xenstore *path* and traverse all entries recursively starting from the provided *path*. The callback provided by user is called for each Xenstore entry.

Then this xss_list_traverse() API is used to rework "xs ls" Xenstore shell command, so it can provide now proper and informative result by dumping all Xenstore entries with values recursively.

Full Xenstore dump example:

```
uart:~$ xs ls /
0: root
1:   tool
2:     xenstored = 
1:   local
2:     domain
3:       0
4:         memory
5:           static-max = 0
5:           target = 0
5:           videoram = -1
4:         control = 
5:           platform-feature-multiprocessor-suspend = 1
5:           platform-feature-xs_reset_watches = 1
5:           shutdown = 
5:           feature-poweroff = 
5:           feature-reboot = 
5:           feature-suspend = 
5:           sysrq = 
4:         vm = 00000000-0000-0000-0000-000000000000
4:         name = Domain-0
4:         domid = 0
4:         device
5:           vbd = 
5:           suspend
6:             event-channel = 
4:         data = 
4:         drivers = 
4:         feature = 
4:         attr = 
4:         error =                                            
3:       1                                                        
4:         cpu                                                  
5:           0                                            
6:             availability = online         
4:         memory
5:           static-max = 524560
5:           target = 524560
5:           videoram = -1
4:         control = 
5:           platform-feature-multiprocessor-suspend = 1
5:           platform-feature-xs_reset_watches = 1
5:           shutdown = 
5:           feature-poweroff = 
5:           feature-reboot = 
5:           feature-suspend = 
5:           sysrq = 
4:         vm = 00000000-0000-0000-0000-000000000001
4:         name = zephyr-1
4:         domid = 1
4:         device
5:           vbd = 
5:           suspend
6:             event-channel = 
4:         data = 
4:         drivers = 
4:         feature = 
4:         attr = 
4:         error = 
1:   vm
2:     00000000-0000-0000-0000-000000000000
3:       name = Domain-0
3:       start_time = 0
3:       uuid = 00000000-0000-0000-0000-000000000000
2:     00000000-0000-0000-0000-000000000001
3:       name = zephyr-1
3:       start_time = 0
3:       uuid = 00000000-0000-0000-0000-000000000001
1:   libxl
2:     0
3:       dm-version = qemu_xen_traditional
3:       type = pvh
2:     1
3:       dm-version = qemu_xen_traditional
3:       type = pvh
1:   fuck_in
2:     ring-ref = 32767
2:     event-channel = 2
1:   fuck_out
2:     ring-ref = 32766
2:     event-channel = 3
```

### DIFF
--- a/xen-shell-cmd/src/xenstore_cmds.c
+++ b/xen-shell-cmd/src/xenstore_cmds.c
@@ -48,30 +48,36 @@ static int xs_write(const struct shell *shell, size_t argc, char **argv)
 	return rc;
 }
 
+static void xs_ls_cb(void *data, const char *key, const char *value, int depth)
+{
+	const struct shell *shell = data;
+
+	if (!key) {
+		shell_print(shell, "%d: root", depth);
+		return;
+	}
+
+	if (value) {
+		shell_print(shell, "%d: %*s%s = %s", depth, depth * 2, " ", key, value);
+	} else {
+		shell_print(shell, "%d: %*s%s", depth, depth * 2, " ", key);
+	}
+}
+
 static int xs_ls(const struct shell *shell, size_t argc, char **argv)
 {
-	int i, len;
-	char **paths;
+	int rc;
 
 	if (argc != 2) {
 		return -EINVAL;
 	}
 
-	paths = xss_list_entries(argv[1], &len);
-
-	if (!paths) {
-		shell_error(shell, "Failed to list xenstore path %s", argv[1]);
-		return -EINVAL;
+	rc = xss_list_traverse(argv[1], xs_ls_cb, (struct shell *)shell);
+	if (rc) {
+		shell_error(shell, "Failed to list xenstore path %s (%d)", argv[1], rc);
 	}
 
-	shell_print(shell, "Listing xenstore path %s", argv[1]);
-	for (i = 0; i < len; i++) {
-		shell_print(shell, "%s", paths[i]);
-		k_free(paths[i]);
-	}
-	k_free(paths);
-
-	return 0;
+	return rc;
 }
 
 static int xs_rm(const struct shell *shell, size_t argc, char **argv)
--- a/xenstore-srv/include/xss.h
+++ b/xenstore-srv/include/xss.h
@@ -97,15 +97,6 @@ int xss_set_perm(const char *path, domid_t domid, enum xs_perm perm);
 int xss_rm(const char *path);
 
 /**
- * List all entries in a directory.
- *
- * @param path Xenstore path
- * @param len Number of entries in the directory
- * @return Dynamically allocated array of dynamically allocated strings with directory entries
- */
-char **xss_list_entries(const char *path, int *len);
-
-/**
  * @brief Xenstore traverse callback
  *
  * @param[in] data User data passed in xss_list_traverse()
--- a/xenstore-srv/include/xss.h
+++ b/xenstore-srv/include/xss.h
@@ -105,6 +105,33 @@ int xss_rm(const char *path);
  */
 char **xss_list_entries(const char *path, int *len);
 
+/**
+ * @brief Xenstore traverse callback
+ *
+ * @param[in] data User data passed in xss_list_traverse()
+ * @param[in] key Xenstore entry name
+ * @param[in] value Xenstore entry value, NULL if not set
+ * @param[in] depth Xenstore tree current depth, starting from 0 (root)
+ */
+typedef void (*xss_traverse_callback_t)(void *data, const char *key, const char *value, int depth);
+
+/**
+ * @brief traverse all entries in a directory recursively
+ *
+ * This function traverses all Xenstore entries starting from @p path and calls
+ * user callback @p cb for each entry.
+ *
+ * @param[in] path Xenstore path
+ * @param[in] cb traverse callback
+ * @param[in] data to be passed in @p cb
+ *
+ * @retval 0 If successful
+ * @retval -EINVAL if @cb not provided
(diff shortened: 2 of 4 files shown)


xenstore-srv/src/xenstore_srv.c
```diff
@@ -1187,6 +1187,52 @@ char **xss_list_entries(const char *path, int *len)
 	return NULL;
 }
 
+static void xss_list_traverse_entry(struct xs_entry *entry, xss_traverse_callback_t cb, void *data,
```
lorc, line 1190: We need to limit depth somehow, because malicious guest can create deeply nested tree and this will blow up stack in Zephyr.

xen-shell-cmd/src/xenstore_cmds.c
```diff
@@ -48,30 +48,36 @@ static int xs_write(const struct shell *shell, size_t argc, char **argv)
 	return rc;
 }
 
-static int xs_ls(const struct shell *shell, size_t argc, char **argv)
+void xs_traverse_cb(void *data, const char *key, const char *value, int depth)
```
lorc, line 51: `static`? Also, maybe call it `xs_ls_cb` to denote it function?

xenstore-srv/include/xss.h
```diff
@@ -105,6 +105,33 @@ int xss_rm(const char *path);
...
+typedef void (*xss_traverse_callback_t)(void *data, const char *key, const char *value, int depth);
+
+/**
+ * @brief traverse all entries in a directory recursively
+ *
+ * This function traverses all Xenstore entries starting from @p path and calls
+ * user callback @p cb for each entry.
+ *
+ * @param[in] path Xenstore path
+ * @param[in[ cb traverse callback
```
Deedone, line 125: Wrong orientation of brackets

xenstore-srv/src/xenstore_srv.c
GrygiriiS: I've been thinking about it - First, I'd add a note to commit:

"NOTE. For a big deeply nested Xenstore tree the shell thread stack CONFIG_SHELL_STACK_SIZE might need to be adjusted"

Second, sure I can add limit, but i'd appreciated if you can advise its value (16, ok?) and how to set it - define or Kconfig?

xenstore-srv/include/xss.h
GrygiriiS: fixed

xen-shell-cmd/src/xenstore_cmds.c
GrygiriiS: fixed

xenstore-srv/src/xenstore_srv.c
lorc: I believe it should be limited via Kconfig. As for default value, yes, I believe 16 will be more than enough, taking into account that there should be no more that 8 levels during normal  use.

xenstore-srv/src/xenstore_srv.c
```diff
@@ -1187,6 +1187,50 @@ char **xss_list_entries(const char *path, int *len)
...
+	struct xs_entry *entry, *iter, *next;
+	sys_dlist_t *children;
+	int depth = 0;
+
+	if (!cb) {
+		return -EINVAL;
+	}
+
+	k_mutex_lock(&xsel_mutex, K_FOREVER);
+	entry = key_to_entry_check_perm(path, 0, XS_PERM_READ);
```
lorc, line 1165: You are checking for XS_PERM_READ only at the top level, but not below.

Taking into account that this is xenstore server, which runs in Dom0, we can ignore access rights altogether. 

As a side note, we need ability to use xenstore commands in other domains, but this is a separate task, I'll create an issue.

GrygiriiS: > You are checking for XS_PERM_READ only at the top level, but not below.

Yep. It's just as is from old list API. For me work it was very important to see Xstore content.

> 
> Taking into account that this is xenstore server, which runs in Dom0, we can ignore access rights altogether.

Sry, not so familiar with XS code yet - so could you clarify "todo" here, pls?

> 
> As a side note, we need ability to use xenstore commands in other domains, but this is a separate task, I'll create an issue.

Checked that at least xenstore-ls and xenstore-list are working for local/domain/x/

lorc: > Sry, not so familiar with XS code yet - so could you clarify "todo" here, pls?

Well, you can remove this check all together, taking into account that this code runs only in Dom0 and Dom0 has access to all entries.

 > Checked that at least xenstore-ls and xenstore-list are working for local/domain/x/

It will not work if we'll run Zephyr as DomU.

GrygiriiS: Ah ok. Target OS was not clear from your comment

xenstore-srv/src/xenstore_srv.c
GrygiriiS: done

xenstore-srv/src/xenstore_srv.c
GrygiriiS: done.

firscity (approved):
@Deedone please take another look

Deedone:
`Reviewed-by: Mykyta Poturai <mykyta_poturai@epam.com>`

Issue #71, opened 2024-04-17
Domain name parsing in shell commands


lorc:
Shell commands should accept both domain ID and domain name. E.g.

```
xu console DomD
```

and 

```
xu console 1
``` 

Should do the same (providing that DomD have id 1)

GrygiriiS:
In my opinion It's overhead from Zephyr point of view. What should be there is smth like "xu list" which should list name/xen_id/<some info>. Then ids used in all other commands.

<some info> : <state ::= "created" | "running" | "paused";> 

From Zephyr point of view there should be always static (configurable) number of supported domains and it's probably time to strictly define domainID/domainNum from Zephyr point of view for cmdline. Of course, It potentially might be possible to have static and dynamic domains (static - configured statically, dynamic - dominfo obtained from some source), but over all number of domains will always be static.

lorc:
We already have AOS use case where they download domains over internet and run them. In this case number of domains is practically unbound. It is limited only by number of free resources.

GrygiriiS:
I think no need to play with dom names and better (a) use numbering not connected to Xen domain id (b) add "list" command to obtain Xen domain id.

lorc:
What is the purpose of the separate domain number then? 

I want interface to resemble used in "xl" tool as it is already familiar for almost everyone working with Xen.

GrygiriiS:
> What is the purpose of the separate domain number then?
> 
> I want interface to resemble used in "xl" tool as it is already familiar for almost everyone working with Xen.

I meant (a) or (b).

 If you want the same functionality as in XEN Linux dom0 - use linux. If you port all of it in Zephyr - you'll get the same problems and disadvantages as in the Linux dom0, so what will be the sense of doing it. 
And xen console here definitely not a product feature.

lorc:
It is not a product feature, but we, as developers, spending too much time in the console. So it is in our interest to have convenient console interface. 

Anyways, I agree with you. We need `xu list` command. I am surprised that it is not implemented yet, actually.

GrygiriiS:
not only 'xu list', but also some way to retrieve and maintain current xen state. Here I'm referring to domains created not by dom0 (dom0less boot), which we might/might not wanna control, but we definitely want console access in case "hvc0" is used.

Some summary:
- dom ctrl commands should accept xen_dom_id
- xen_dom_id is known or can be obtained through `xu list`

potential addon:
[...]

PR #82, opened 2024-05-07
Allow to use domain name instead of domid in commands

Deedone:
Add the ability to use domain name instead of domid in the xu shell
commands. If the passed identifier is all numbers, then it is
considered as domid, otherwise, it is considered as domain name.
Updated commands description to reflect this change.

Also add the command to list domain infos so we can actually know the domain names.
Fixes https://github.com/xen-troops/zephyr-xenlib/issues/71

### DIFF
--- a/xen-dom-mgmt/src/xen-dom-mgmt.c
+++ b/xen-dom-mgmt/src/xen-dom-mgmt.c
@@ -6,6 +6,7 @@
 
 #include <zephyr/sys/byteorder.h>
 #include <zephyr/xen/dom0/domctl.h>
+#include <zephyr/xen/dom0/sysctl.h>
 #include <zephyr/xen/hvm.h>
 #include <zephyr/logging/log.h>
 
@@ -619,6 +620,18 @@ struct xen_domain_cfg *domain_find_config(const char *name)
 #endif
 
 	return NULL;
+}
+
+int get_domain_name(unsigned short domain_id, char *name, int len)
+{
+#ifdef CONFIG_XEN_STORE_SRV
+	char path[sizeof("/local/domain/32768/name")];
+
+	snprintf(path, sizeof(path), "/local/domain/%u/name", domain_id);
+	return xss_read(path, name, len);
+#else
+	return -EINVAL;
+#endif
 }
 
 __weak int domain_get_user_cfg_count(void)
--- a/xen-dom-mgmt/src/xen-dom-mgmt.c
+++ b/xen-dom-mgmt/src/xen-dom-mgmt.c
@@ -634,6 +634,32 @@ int get_domain_name(unsigned short domain_id, char *name, int len)
 #endif
 }
 
+uint32_t find_domain_by_name(char *arg)
+{
+	char domname[CONTAINER_NAME_SIZE];
+	struct xen_domctl_getdomaininfo infos[CONFIG_DOM_MAX];
+	uint32_t domid = 0;
+	int i, ret;
+
+	ret = xen_sysctl_getdomaininfo(infos, 0, CONFIG_DOM_MAX);
+	if (ret < 0) {
+		goto out;
+	}
+
+	for (i = 0; i < ret; i++) {
+		if (!get_domain_name(infos[i].domain, domname,
+					    CONTAINER_NAME_SIZE)) {
+			if (strncmp(domname, arg, CONTAINER_NAME_SIZE) == 0) {
+				domid = infos[i].domain;
+				break;
+			}
+		}
+	}
+
+out:
+	return domid;
+}
+
 __weak int domain_get_user_cfg_count(void)
 {
 	return 0;
--- a/xen-dom-mgmt/src/xen-dom-mgmt.c
+++ b/xen-dom-mgmt/src/xen-dom-mgmt.c
@@ -4,6 +4,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#undef _POSIX_C_SOURCE
+#define _POSIX_C_SOURCE 200809L
+
 #include <zephyr/sys/byteorder.h>
 #include <zephyr/xen/dom0/domctl.h>
 #include <zephyr/xen/dom0/sysctl.h>
@@ -47,6 +50,7 @@ static int dom_num = 0;
 
 static sys_dlist_t domain_list = SYS_DLIST_STATIC_INIT(&domain_list);
 K_MUTEX_DEFINE(dl_mutex);
+K_MUTEX_DEFINE(create_mutex);
 
 static void arch_prepare_domain_cfg(struct xen_domain_cfg *dom_cfg,
 				    struct xen_arch_domainconfig *arch_cfg)
@@ -678,6 +682,7 @@ int domain_create(struct xen_domain_cfg *domcfg, uint32_t domid)
 	struct vcpu_guest_context vcpu_ctx;
 	struct xen_domain *domain;
 	struct modules_address modules = {0};
+	char *name;
 
 	if (dom_num >= CONFIG_DOM_MAX) {
 		LOG_ERR("Runtime exceeds maximum number of domains");
@@ -700,8 +705,14 @@ int domain_create(struct xen_domain_cfg *domcfg, uint32_t domid)
 	}
 	memset(domain, 0, sizeof(*domain));
 	domain->domid = domid;
+	/* Fallback to name if domain_name is not set */
+	if (strnlen(domcfg->domain_name, CONTAINER_NAME_SIZE) > 0) {
+		name = domcfg->domain_name;
+	} else {
+		name = domcfg->name;
+	}
 
-	snprintf(domain->name, CONTAINER_NAME_SIZE, "%s", domcfg->name);
+	snprintf(domain->name, CONTAINER_NAME_SIZE, "%s", name);
 	rc = xen_domctl_max_vcpus(domid, domcfg->max_vcpus);
 	if (rc) {
 		LOG_ERR("Failed to set max vcpus for domain#%u (rc=%d)", domid, rc);
@@ -792,7 +803,17 @@ int domain_create(struct xen_domain_cfg *domcfg, uint32_t domid)
 	}
 #endif
 
+	k_mutex_lock(&create_mutex, K_FOREVER);
+	if (find_domain_by_name(name) != 0) {
+		rc = -EEXIST;
+		LOG_ERR("Domain with name %s already exists", name);
+		k_mutex_unlock(&create_mutex);
+		goto stop_domain_console;
+	}
+
 	rc = xs_initialize_xenstore(domid, domain);
+	k_mutex_unlock(&create_mutex);
+
 	if (rc) {
 		goto stop_domain_console;
 	}
--- a/xen-shell-cmd/src/xen_cmds.c
+++ b/xen-shell-cmd/src/xen_cmds.c
@@ -6,6 +6,7 @@
 
 #include <zephyr/shell/shell.h>
 #include <zephyr/logging/log.h>
+#include <zephyr/xen/dom0/sysctl.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -15,6 +16,7 @@
 #endif
 
 LOG_MODULE_REGISTER(xen_shell);
+static struct xen_domctl_getdomaininfo infos[CONFIG_DOM_MAX];
 
 static struct xen_domain_cfg domcfg_tmp;
 
@@ -264,6 +266,41 @@ int xen_config_list(const struct shell *shell, size_t argc, char **argv)
 
 	return 0;
 }
+static int domain_list(const struct shell *shell, size_t argc, char **argv)
+{
+	int i, ret;
+	char domname[CONTAINER_NAME_SIZE];
+	unsigned long memkb;
+
+	ARG_UNUSED(argc);
+	ARG_UNUSED(argv);
+
+	ret = xen_sysctl_getdomaininfo(infos, 0, CONFIG_DOM_MAX);
+	if (ret < 0) {
+		goto out;
+	}
+	shell_fprintf(shell, SHELL_NORMAL,
+			"Name                                        ID   Mem VCPUs\tState\tTime(s)\n");
+	for (i = 0; i < ret; i++) {
+		if (get_domain_name(infos[i].domain, domname, CONTAINER_NAME_SIZE)) {
+			strncpy(domname, "(NULL)", CONTAINER_NAME_SIZE);
+		}
+		memkb = infos[i].tot_pages * XEN_PAGE_SIZE / 1024;
+		shell_fprintf(shell, SHELL_NORMAL, "%-40s %5d %5lu %5d      %c%c%c%c%c  %8.1f\n",
+					  domname,
+					  infos[i].domain,
+					  (unsigned long) (memkb / 1024),
+					  infos[i].nr_online_vcpus,
+					  infos[i].flags & XEN_DOMINF_running ? 'r' : '-',
+					  infos[i].flags & XEN_DOMINF_blocked ? 'b' : '-',
+					  infos[i].flags & XEN_DOMINF_paused ? 'p' : '-',
+					  infos[i].flags & XEN_DOMINF_shutdown ? 's' : '-',
+					  infos[i].flags & XEN_DOMINF_dying ? 'd' : '-',
+					  ((double)infos[i].cpu_time / 1e9));
+	}
+out:
+	return ret;
+}
 
 SHELL_STATIC_SUBCMD_SET_CREATE(
 	subcmd_xu,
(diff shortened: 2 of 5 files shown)


xen-shell-cmd/src/xen_cmds.c
```diff
@@ -34,6 +38,18 @@ uint32_t parse_domid(size_t argc, char **argv)
 	return 0;
 }
 
+static int get_domain_name(unsigned short domain_id, char *name, int len)
+{
+#ifdef CONFIG_XEN_STORE_SRV
```
GrygiriiS, line 43: Hm. Why?the domain name is part of struct xen_domain ?

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -132,6 +148,51 @@ int domu_unpause(const struct shell *shell, size_t argc, char **argv)
...
+{
+	struct xen_domctl_getdomaininfo *infos;
+	int i, ret;
+	char domname[CONTAINER_NAME_SIZE];
+	unsigned long memkb;
+
+	ARG_UNUSED(argc);
+	ARG_UNUSED(argv);
+
+	infos = k_malloc(sizeof(*infos) * CONFIG_DOM_MAX);
```
GrygiriiS, line 208: In any case no more k_malloc()!

xen-shell-cmd/src/xstat_cmds.c
```diff
@@ -107,30 +38,6 @@ static int xstat_shell_stat(const struct shell *shell, size_t argc, char **argv)
 	shell_print(shell, "CPU freq(kHz): %lld", stat.cpu_hz / 1000);
 	shell_print(shell, "Total mem(K) : %llu", stat.tot_mem / 1024);
 	shell_print(shell, "Free mem(K)  : %llu", stat.free_mem / 1024);
```
GrygiriiS, line 40: Logically it shell or dropped completely or changed to smth. like "xstat sysinfo"

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -132,6 +148,51 @@ int domu_unpause(const struct shell *shell, size_t argc, char **argv)
...
+		goto out;
+	}
+	ret = xen_sysctl_getdomaininfo(infos, 0, CONFIG_DOM_MAX);
+	if (ret < 0) {
+		goto out_free;
+	}
+	shell_fprintf(shell, SHELL_NORMAL,
+			"Name                                        ID   Mem VCPUs\tState\tTime(s)\n");
+	for (i = 0; i < ret; i++) {
+		if (get_domain_name(infos[i].domain, domname,
```
GrygiriiS, line 220: Why not to traverse domain_list instead of playing with Xenstore, we should not have unaccounted domains.
For each domain xen_domctl_getdomaininfo() can be called.

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -50,6 +50,53 @@ static int get_domain_name(unsigned short domain_id, char *name, int len)
...
+	char domname[CONTAINER_NAME_SIZE];
+	struct xen_domctl_getdomaininfo *infos;
+	uint32_t domid = 0;
+	int i, ret;
+
+	if (is_all_numbers(arg)) {
+		return atoi(arg);
+	}
+
+	infos = k_malloc(sizeof(*infos) * CONFIG_DOM_MAX);
```
GrygiriiS, line 77: No more! No more k_malloc()! It's not Linux user space.

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -50,6 +50,53 @@ static int get_domain_name(unsigned short domain_id, char *name, int len)
...
+
+	ret = xen_sysctl_getdomaininfo(infos, 0, CONFIG_DOM_MAX);
+	if (ret < 0) {
+		goto out_free;
+	}
+
+	for (i = 0; i < ret; i++) {
```
GrygiriiS, line 84: Why such a complexity!? Why do not use our domain_list?

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -50,6 +50,53 @@ static int get_domain_name(unsigned short domain_id, char *name, int len)
...
+}
+
+static uint32_t parse_domid_or_name(char *arg)
+{
+	char domname[CONTAINER_NAME_SIZE];
+	struct xen_domctl_getdomaininfo *infos;
+	uint32_t domid = 0;
+	int i, ret;
+
+	if (is_all_numbers(arg)) {
```
GrygiriiS, line 73: I vote for additional explicit option, like '-n'.

Or question - why I can't have domain name "1234"? ;)

Deedone: We try to keep the interface similar to the xl tool.

xen-shell-cmd/src/xen_cmds.c
Deedone: Care to explain why?

xen-shell-cmd/src/xen_cmds.c
Deedone: Because info in the xenstore is 100% valid, in my opinion domain_list should not contain names at all.

xen-shell-cmd/src/xen_cmds.c
lorc: Well, this is how xen tools do this. In case of xen tools, they don't have persistent list of domains. List of domains is managed by Xen itself + there is additional information in XenStore.  This code copies this behavior, but I am not sure that this is the right way, to be honest.

xen-shell-cmd/src/xen_cmds.c
lorc: Are you proposing to allocate items on stack?

xen-shell-cmd/src/xen_cmds.c
lorc: Well, if we want to mimic original xen tools behavior - then yes. But I am wondering - are we really want?

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -50,6 +50,53 @@ static int get_domain_name(unsigned short domain_id, char *name, int len)
 #endif
 }
 
+static bool is_all_numbers(char *str)
+{
+	int i;
+
+	for (i = 0; str[i] != '\0'; i++) {
+		if (str[i] < '0' || str[i] > '9') {
```
lorc, line 58: So, zephyr does not provide `isdigit()`?

xen-shell-cmd/src/xen_cmds.c
lorc: Yeah, I am not sure that xen tools allow all-digits domain name. I believe it is not and we need the same check. 

It is okay in this case, because here we can use explicit option. But problems will arise in other parts, like PV driver configuration, where you can use either domain name or domain id.

xen-shell-cmd/src/xen_cmds.c
GrygiriiS: Yeah. Lets see before:
- "xstat stat" shows us Xen state. in general the same can be achieved with xen console [q, d, e]
- no information about Dom0 domain list can be shown
Now:
- "xstat stat" truncated, and functionality moved to "xu list" and shows Xen state now
- still no info about Dom0 actual domain_list state which could be differ due to zombies at least

It seems, xstat can be left as is and "xu list" can show Dom0 actual status. That seems will give us full system overview (including no Xen console case). Note. In Zephyr SHELL is NOT management, production interface - it's debug interface first of all.

xen-shell-cmd/src/xen_cmds.c
GrygiriiS: see above

xen-shell-cmd/src/xen_cmds.c
GrygiriiS: First of all - As you explained to me there could be zombies domains as result Xen can have more domains that we actually support in Zephyr Dom0. Which means if there are CONFIG_DOM_MAX=4 and Xen domains [0, 1, z, z, 4, 5] this will not show info about domains [4, 5] while they are still in domain_list and functional.

In my opinion the domain_list state has to be shown here.

And yes xen_domctl_getdomaininfo info; + xen_domctl_getdomaininfo() in case of traversing domain_list.

xen-shell-cmd/src/xen_cmds.c
GrygiriiS: There is one more thing - Dom0less domain might not have names or even xenstore, so if dom0less option not enabled it might show garbage.

xen-shell-cmd/src/xen_cmds.c
GrygiriiS: ok. if it will be decided to continue with Xen data instead of displaying our domain_list info - In my opinion simplest way would be to have struct xen_domctl_getdomaininfo  infos[X], where X=4 for example. Then continuously request xen_sysctl_getdomaininfo() until it return `ret<X`.

static alloc is safe as only one SHELL cmd can be executed at time.

GrygiriiS:
Could you share example output for:
- patch 1 "xen_cmds: Add domain list command"
- patch 2 "xstat: Remove domain listing from xstat command"

GrygiriiS:
Huh. Sorry but this required more work or at least testing. Below sequence will create 3 domu all with the same name:
```
uart:~$ xu config_list 
rpi_5_domd
rpi_5_domu
uart:~$ xu create rpi_5_domu
uart:~$ xu create rpi_5_domu
uart:~$ xu create rpi_5_domu
W: Domain device tree generation is not supported
I: rambase = 40000000, ramsize = 16777216
I: kernbase = 40000000 kernsize = 364548, dtbsize = 11058
I: kernsize_aligned = 2097152
I: DTB will be placed on addr = 0x40e00000
I: Created domain is paused
To unpause issue: xu unpause 4
domain:0 created

uart:~$ xstat stat 
XEN version 4.19-unstable
CPUs         : 4
CPU freq(kHz): 54000
Total mem(K) : 8384512
Free mem(K)  : 7536972
Domain #  0     : Domain-0
State           : -----r
CPU ns          : 2565065041
VCPUs           : 1
MEM reserved(K) : 131072
MEM allowed(K)  : 131072
SSID            : 3
---------------------------------
Domain #  1     : Dom0less-1
State           : -----r
CPU ns          : 43957473556
VCPUs           : 1
MEM reserved(K) : 524304
MEM allowed(K)  : 524304
SSID            : 6
---------------------------------
Domain #  2     : rpi_5_domu
State           : ----p-
CPU ns          : 0
VCPUs           : 1
MEM reserved(K) : 16400
MEM allowed(K)  : 163840
SSID            : 6
---------------------------------
Domain #  3     : rpi_5_domu
State           : ----p-
CPU ns          : 0
VCPUs           : 1
MEM reserved(K) : 16400
MEM allowed(K)  : 163840
SSID            : 6
---------------------------------
Domain #  4     : rpi_5_domu
State           : ----p-
CPU ns          : 0
VCPUs           : 1
MEM reserved(K) : 16400
MEM allowed(K)  : 163840
SSID            : 6
---------------------------------
```

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -704,6 +704,12 @@ int domain_create(struct xen_domain_cfg *domcfg, uint32_t domid)
 	memset(&config, 0, sizeof(config));
 	prepare_domain_cfg(domcfg, &config);
 	config.grant_opts = XEN_DOMCTL_GRANT_version(1);
+
+	if (find_domain_by_name(domcfg->name) != 0) {
```
lorc, line 708: This is a weak check unless `domain_create` is not called in critical section. 

There can be race with two `domain_create` calls in separate thread

Deedone: Added lock

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -704,6 +705,7 @@ int domain_create(struct xen_domain_cfg *domcfg, uint32_t domid)
 	memset(&config, 0, sizeof(config));
 	prepare_domain_cfg(domcfg, &config);
 	config.grant_opts = XEN_DOMCTL_GRANT_version(1);
+
```
lorc, line 708: Stray change

lorc (approved):
With the last small nit addressed:

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -809,7 +810,17 @@ int domain_create(struct xen_domain_cfg *domcfg, uint32_t domid)
 	}
 #endif
 
+	k_mutex_lock(&create_mutex, K_FOREVER);
+	if (find_domain_by_name(domcfg->name) != 0) {
+		rc = -EEXIST;
```
GrygiriiS, line 815: Sorry, but absolutely **do not** agree with this limitation.

1. domcfg->name is first of all **configuration name**, also provided to the user by `xu config_list`, and the fact that it is used now as domain name doesn't mean it can be used as reject factor here.
2. There are absolutely no reasons to block creation of multiple domains using the same configuration - domains can base their services/jobs depending on remote configuration obtained through network as an example.

This has to be solved in different way by guarantee domain name is unique during creation. The domcfg->name could be a base and extended with Xen domain_id, for example.

lorc: > Sorry, but absolutely do not agree with this limitation.

Well, we require domain name to be unique. So we need this limitation.

But I agree with you in sense that we should not use `domcfg->name` as a domain name, because this is a configuration name.

We need a separate field/parameter `domain_name`.

lorc: It can work, yes. Or we can do what  Xen toolstack does: use domain name from config file by default and allow a user to override it via command line.

But, frankly, this is quite complex logic for our current needs. So, maybe let's stick to your idea? @firscity, what is your opinion?

firscity: In my opinion we should use new `domcfg->domain_name` because we need to know domain name when will create complex configs (e.g. with backend configuration), usage of domid will made them dependent on runtime.

Also, I don't see much sense in using same configuration for multiple domains, because it will work only for simple domains without any e.g interrupts (you can not add same interrupt to few domains).

GrygiriiS: ok

Deedone:
Added separate domain_name field

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -264,6 +266,41 @@ int xen_config_list(const struct shell *shell, size_t argc, char **argv)
...
+	}
+	shell_fprintf(shell, SHELL_NORMAL,
+			"Name                                        ID   Mem VCPUs\tState\tTime(s)\n");
+	for (i = 0; i < ret; i++) {
+		if (get_domain_name(infos[i].domain, domname, CONTAINER_NAME_SIZE)) {
+			strcpy(domname, "(NULL)");
+		}
+		memkb = infos[i].tot_pages * XEN_PAGE_SIZE / 1024;
+		shell_fprintf(shell, SHELL_NORMAL, "%-40s %5d %5lu %5d      %c%c%c%c%c  %8.1f\n",
+						domname,
```
firscity, line 313: Strange formatting, please, place these parameters on same level with `shell_fprintf(` brace.

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -264,6 +266,41 @@ int xen_config_list(const struct shell *shell, size_t argc, char **argv)
...
+
+	ret = xen_sysctl_getdomaininfo(infos, 0, CONFIG_DOM_MAX);
+	if (ret < 0) {
+		goto out;
+	}
+	shell_fprintf(shell, SHELL_NORMAL,
+			"Name                                        ID   Mem VCPUs\tState\tTime(s)\n");
+	for (i = 0; i < ret; i++) {
+		if (get_domain_name(infos[i].domain, domname, CONTAINER_NAME_SIZE)) {
+			strcpy(domname, "(NULL)");
```
firscity, line 309: Please, use limited version of `str` functions (`strncmp`, `strncpy` etc.), this relates to whole PR. String size here is some define in header file, that may be changed at some moment and cause unexpected issues.

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -633,6 +633,32 @@ int get_domain_name(unsigned short domain_id, char *name, int len)
...
+
+	ret = xen_sysctl_getdomaininfo(infos, 0, CONFIG_DOM_MAX);
+	if (ret < 0) {
+		goto out;
+	}
+
+	for (i = 0; i < ret; i++) {
+		if (!get_domain_name(infos[i].domain, domname,
+					    CONTAINER_NAME_SIZE)) {
+			if (strcmp(domname, arg) == 0) {
```
firscity, line 652: `strncmp()`

firscity (commented):
With small comments fixed:

@GrygiriiS @lorc please, re-check

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -701,7 +702,7 @@ int domain_create(struct xen_domain_cfg *domcfg, uint32_t domid)
 	memset(domain, 0, sizeof(*domain));
 	domain->domid = domid;
 
-	snprintf(domain->name, CONTAINER_NAME_SIZE, "%s", domcfg->name);
+	snprintf(domain->name, CONTAINER_NAME_SIZE, "%s", domcfg->domain_name);
```
lorc, line 705: This will break existing code that don't know about `domain_name`. You need to add fallback here: if `domain_name` is empty - then use `name`.

lorc (approved):

PR #83, opened 2024-05-07
Xenlib misc updates and fixes

GrygiriiS:
Set of misc, not critical updates and fixes.

### DIFF
--- a/xen-dom-mgmt/src/xen-dom-mgmt.c
+++ b/xen-dom-mgmt/src/xen-dom-mgmt.c
@@ -1044,26 +1044,15 @@ static int dom0less_init(void)
 
 static int init_domain0(void)
 {
+	struct xen_domctl_getdomaininfo dominfo;
 	int ret = 0;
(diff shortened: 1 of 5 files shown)


xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -1044,26 +1044,15 @@ static int dom0less_init(void)
 
 static int init_domain0(void)
 {
+	struct xen_domctl_getdomaininfo dominfo;
```
lorc, line 1047: Taking into account size of `xen_domctl_getdomaininfo` it is debatable if we should store it in the stack. Better to have -ENOMEM that stack overrun.

lorc (approved):
Thanks

lorc:
Oh, and one more thing. Please configure spell checker in your editor, so we can avoid things like "but there is
**mor** simple" in commit messages.

lorc:
@firscity, could you please merge this?

PR #85, opened 2024-05-10
xen_console: Fix deadlock in xen_stop_domain_console()

Deedone:
Both xen_stop_domain_console and console thread are trying to acquire global_console_lock and unset current_console at the same time. This can prevent console thread from ever stopping if xen_stop_domain_console acquires the lock first.

This patch fixes the issue by unlocking the global_console_lock before joining the console thread, and leaving the job of unsetting the current console to the console thread.

### DIFF
--- a/xen-console-srv/src/xen_console.c
+++ b/xen-console-srv/src/xen_console.c
@@ -345,10 +345,11 @@ int xen_stop_domain_console(struct xen_domain *domain)
 	/* Stop attached console if any */
 	if (current_console == console) {
 		k_sem_give(&console->int_sem);
+		k_mutex_unlock(&global_console_lock);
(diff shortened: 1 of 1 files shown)


Deedone:
Creating a separate PR as requested in https://github.com/xen-troops/zephyr-xenlib/pull/72

firscity:
IIUC here should be first two patches from #72

Deedone:
As I understand it was requested to specifically separate and merge the fix first, the second patch is a signature change that is only actually useful for refcounting.

firscity (approved):

GrygiriiS:
Thank you. `Reviewed-by: Grygorii Strashko <grygorii_strashko@epam.com>`

lorc:
@firscity please do the honors

PR #87, opened 2024-05-14
xen-dom-mgmt: Use correct domid for domain_post_create

Deedone:
Currently, when domid auto allocation is used, we pass 0 to domain_post_create, causing it to fail. Fix this by returning the allocated domid from domain_create and passing it to domain_post_create.

### DIFF
--- a/xen-dom-mgmt/src/xen-dom-mgmt.c
+++ b/xen-dom-mgmt/src/xen-dom-mgmt.c
@@ -793,7 +793,11 @@ int domain_create(struct xen_domain_cfg *domcfg, uint32_t domid)
 	++dom_num;
 	k_mutex_unlock(&dl_mutex);
 
-	return rc;
+	if (rc) {
+		return rc;
+	}
+
+	return domid;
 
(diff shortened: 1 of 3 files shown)


xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -793,7 +793,11 @@ int domain_create(struct xen_domain_cfg *domcfg, uint32_t domid)
 	++dom_num;
 	k_mutex_unlock(&dl_mutex);
 
-	return rc;
+	if (rc) {
+		return rc;
+	} else {
```
firscity, line 798: nit: else is not actually needed here

firscity (approved):

GrygiriiS:
`Reviewed-by: Grygorii Strashko <grygorii_strashko@epam.com>`

GrygiriiS:
Actually this is not compatible neither with xrun neither with aos where domain_create return value  is checked as !ret

PR #88, opened 2024-05-29
Xenlib update to support fatfs on sdmmc

GrygiriiS:
This series is prerequisite to enable FATFS storage on SDMMC for zephyr-dom0-xt application, instead of default littlefs.

It fixes build failure when littlefs is not enabled.

It adds support for DMA debounce buffer in xrun/storage which is required to enable DMA for SDMMC device, because xenlib loads Kernel image directly in foreign guest domain memory which can be used with DMA.

It adds  option to read domain partial device-tree binary from storage with "xu create" shell command.

### DIFF
--- a/xen-shell-cmd/src/xen_cmds.c
+++ b/xen-shell-cmd/src/xen_cmds.c
@@ -16,6 +16,57 @@
 
 LOG_MODULE_REGISTER(xen_shell);
 
+#if defined(CONFIG_XEN_DOMCFG_READ_PDT)
+
+static uint8_t pfdt_read_buf[CONFIG_PARTIAL_DEVICE_TREE_SIZE] __aligned(8);
+
+static int xen_cmd_read_pfdt(struct xen_domain_cfg *domcfg)
+{
+	size_t pfdt_size;
+	int ret;
+
+	if (domcfg->dtb_start && domcfg->dtb_end) {
+		return 0;
+	}
+
+	if (!domcfg->image_dt_get_size && !domcfg->image_dt_read) {
+		LOG_DBG("PDT read callback not provided");
+		return 0;
+	}
+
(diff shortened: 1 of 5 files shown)


xen-shell-cmd/src/xen_cmds.c
```diff
@@ -71,6 +122,13 @@ static int domu_create(const struct shell *shell, int argc, char **argv)
 
 	parse_and_fill_flags(argc, argv, cfg);
 
+#if defined(CONFIG_XEN_DOMCFG_READ_PDT)
+	ret = xen_cmd_read_pfdt(cfg);
+	if (ret) {
```
lorc, line 127: Partial DTB is completely optional feature. It may not be present at all.

GrygiriiS: I'll update below to return 0. ok?
```
	if (!domcfg->image_dt_get_size && !domcfg->image_dt_read) {
		LOG_ERR("PDT read callbakc not provided");
		return -ENOTSUP;
	}
```
Then if no (dtb_start/dtb_end) and no (image_dt_get_size/image_dt_read) then it will satisfy optional PDT choice.

GrygiriiS:
checkpatch warning is false positive.

xrun/Kconfig
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) 2019 Peter Bigot Consulting, LLC
```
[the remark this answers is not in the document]
GrygiriiS: fixed

xen-shell-cmd/src/xen_cmds.c
GrygiriiS: updated

lorc (approved):
Thank you

PR #90, opened 2024-06-03
xen-dom-mgmt: Remove pages from physmap when unmapping

Deedone:
When unmapping a region, pages should be removed from the self physmap as well. This is necessary to avoid leaking page references and ensure that domains can be properly destroyed.

### DIFF
--- a/xen-dom-mgmt/src/mem-mgmt.c
+++ b/xen-dom-mgmt/src/mem-mgmt.c
@@ -102,19 +102,6 @@ uint64_t xenmem_populate_physmap(int domid,
 	return i;
 }
 
-#if defined(CONFIG_XEN_REGIONS)
-static void *get_region_space(uint64_t nr_pages)
-{
-	return xen_region_get_pages(nr_pages);
-}
-
-static int put_region_space(void *mapped_addr, uint64_t nr_pages)
-{
-	xen_region_put_pages(mapped_addr, nr_pages);
-	return 0;
-}
-#else /* CONFIG_XEN_REGIONS */
-
 static uint64_t region_space_remove(void *mapped_addr, uint64_t nr_pages)
 {
 	uint64_t base_pfn, i;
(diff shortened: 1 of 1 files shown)


GrygiriiS:
I tried it and do not see zombies anymore: `Tested-by: Grygorii Strashko <grygorii_strashko@epam.com>`

lorc (commented):
Before I continue reviewing: have you tried this with `CONFIG_XEN_REGIONS=n`?

Deedone:
>have you tried this with CONFIG_XEN_REGIONS=n
@lorc builds and works okay for me

xen-dom-mgmt/src/mem-mgmt.c
```diff
@@ -260,7 +260,18 @@ int xenmem_unmap_region(uint64_t nr_pages, void *mapped_addr)
 		return -EFAULT;
 	}
 #else
+	uint64_t base_pfn = xen_virt_to_gfn(mapped_addr);
+	int rc, i = 0;
+
 	xen_region_unmap(mapped_addr, nr_pages);
+	for (i = 0; i < nr_pages; i++) {
```
firscity, line 267: It seems like `region_space_remove()` that is called in other part of ifdef is doing the same thing
Maybe it can be re-arranged to be called in both cases?

lorc: Yes, thanks. This is what I wanted to asks. It seems that `xendom_remove_from_physmap` needs to be called in both cases but I can't see where is it called in other part of `#if` block. Maybe, some we need some refactoring here...

xen-dom-mgmt/src/mem-mgmt.c
```diff
@@ -102,19 +102,6 @@ uint64_t xenmem_populate_physmap(int domid,
 	return i;
 }
 
-#if defined(CONFIG_XEN_REGIONS)
```
lorc, line 105: Do you really need to move those functions around? Looks like unnecessary change.

lorc (approved):
Looks good apart from that small nitpick.

With the comment addressed: 

`Acked-by: Volodymyr Babchuk <volodymyr_babchuk@epan.com>`

xen-dom-mgmt/src/mem-mgmt.c
Deedone: I actually moved region_space_remove out of the if .. else guards.  For some reason the diff generated in a very strange way. Maybe move it farther?

firscity (approved):

xen-dom-mgmt/src/mem-mgmt.c
lorc: Ah, I see now. No, its fine. Leave it as is.

PR #92, opened 2024-06-14
xen-shell-cmd: ensure shell cmds do not change domain cfg

GrygiriiS:
The commit 7c39b5623258 ("xen-shell-cmd: add optional parameter for pausing domain on creation") and commit e528793c3ccc ("xen-dom-mgmt: add option to read domain partial device-tree") added new functionality to the xenlib shell command interface, but at the same time those commits change some fields in struct xen_domain_cfg object while executing the "xu create" command. This leads to unexpected behavior, like for the first commit:

 domcfg->f_paused = false
 xu create rpi_5_domu    --> unpaused
 xu create rpi_5_domu -p --> paused cfg->f_paused=true
 xu create rpi_5_domu    --> paused cfg->f_paused=true
 xu create rpi_5_domu -p --> paused cfg->f_paused=true
 xu create rpi_5_domu    --> paused cfg->f_paused=true

so once domain is creted with "-p" option - it will be always created in "paused" state after this.

For the second commit situation is similar, partial DT will be red from storage only once.

Hence shell interface allows to execute only one command at time the above issues are fixed by saving and restoring modified fields of struct xen_domain_cfg.

### DIFF
--- a/xen-shell-cmd/src/xen_cmds.c
+++ b/xen-shell-cmd/src/xen_cmds.c
@@ -16,6 +16,8 @@
 
 LOG_MODULE_REGISTER(xen_shell);
 
+static struct xen_domain_cfg domcfg_tmp;
+
 #if defined(CONFIG_XEN_DOMCFG_READ_PDT)
 
 static uint8_t pfdt_read_buf[CONFIG_PARTIAL_DEVICE_TREE_SIZE] __aligned(8);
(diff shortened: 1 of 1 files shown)


xen-shell-cmd/src/xen_cmds.c
```diff
@@ -16,6 +16,16 @@
...
 
+struct xen_shell_dom_mgmt_context {
+	bool f_restore_paused:1;
+	bool f_paused_old:1;
+#if defined(CONFIG_XEN_DOMCFG_READ_PDT)
+	bool f_restore_dtb_ptrs:1;
+#endif /* CONFIG_XEN_DOMCFG_READ_PDT */
+};
+
+struct xen_shell_dom_mgmt_context shell_dom_mgmt_ctx;
```
lorc, line 27: `static`

lorc (commented):
Okay, I see the issue and I agree that shell should not change domain config.

But I am confused a bit  by the solution. IMO, it adds unnecessary complexity. Maybe it would be simpler to create copy of `domain_cfg` and use it? This will ensure that any similar issues will be covered as well.

GrygiriiS:
> Okay, I see the issue and I agree that shell should not change domain config.
> 
> But I am confused a bit by the solution. IMO, it adds unnecessary complexity. Maybe it would be simpler to create copy of `domain_cfg` and use it? This will ensure that any similar issues will be covered as well.

I'll think about and try it. Thanks.

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -119,29 +121,33 @@ static int domu_create(const struct shell *shell, int argc, char **argv)
 		shell_error(shell, "Config %s not found", name);
 		return -EINVAL;
 	}
+	domcfg_tmp = *cfg;
```
firscity, line 124: This part (`domain_create()`) can be called from multiple places, so `domcfg_tmp` may be over-written. You need to use either locks, or complete independent copy of the config.

xen-shell-cmd/src/xen_cmds.c
```diff
@@ -119,29 +121,33 @@ static int domu_create(const struct shell *shell, int argc, char **argv)
...
 	if (ret) {
-		return ret;
+		goto exit_err;
 	}
 
 	shell_info(shell, "domain:%u created", domid);
-	return 0;
+exit_err:
+	/* restore domain cfg */
+	*cfg = domcfg_tmp;
```
firscity, line 149: Are you sure that every provider of `domain_find_config()` will be happy with this? The best way will be using a complete copy and don't touch original config at all

firscity (changes requested):

firscity (approved):

firscity:
@lorc can you please re-check?

lorc (approved):

PR #94, opened 2024-07-02
xen-dom-mgmt: add domains xsm security labels id cfg

GrygiriiS:
The XSM policy is completely user defined, and any IDs associated with policy elements (like user, role and type) are auto-generated. The pretty complex Xen tools are responsible for converting from string "seclabel" (xl.cfg) into SSID used by Xen.

For example, below conversion happens with XSM security label, specified in xl.cfg (very simplified):

seclabel='system_u:system_r:domU_t' => domU => SECINITSID_DOMU => 12

As result, the only available option right now is to allow user to specify domain SSID in struct xen_domain_cfg.

This patch adds possibility to specify XSM security labels id for domain in struct xen_domain_cfg, which than is passed to Xen unchanged during domain creation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,3 +111,39 @@ config XEN_DOMAIN_MANAGEMENT
 config XEN_CONSOLE_SRV
 ```
 For more information on the configuration options, please refer to the `Kconfig` file.
+
+## Xen Security Modules (XSM) - FLASK support (experimental)
+
+The XSM policy is completely user defined, and any IDs associated with policy
+elements (like user, role and type) are auto-generated at Xen build time.
+
+For XSM to work properly the domU need to be marked with security label
+in their configuration. Without this the domains will be classified “unlabeled”.
+
+Example line from a domU configuration (xl.cfg):
+```
+seclabel='system_u:system_r:domU_t'
+```
+
+In Linux world, the pretty complex set of Xen tools is responsible for
+converting from above security label string into numeric SSID value used by
+Xen internally. For example, below conversion happens with XSM domain security
+label, specified in very simplified form:
+```
+    seclabel='system_u:system_r:domU_t' => domU => SECINITSID_DOMU => 12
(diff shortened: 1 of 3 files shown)


GrygiriiS:
@lorc @firscity @otyshchenko1 @oleksiimoisieiev @dsemenets

lorc (commented):
Well, the change itself is very simple, but I fail to understand how it can be used in the real world scenario. 

Okay, now I can `ssidref` to something. But **what** value exactly I should use?

This path itself is good, but it does not solve the issue. I am proposing to:
1. Leave this patch as is
2. Add another patch that documents how and where user can get `ssidref` value
3. Clearly mark the second patch as a temporary solution.
4. Create github issue describing the problem with seclabel assignment so we can find proper solution in due time.

GrygiriiS:
This patch is based mostly on "reverse engineering" approach, and only one thing I know is that SSIDs can be found in xen/xen/xsm/flask/include/flask.h. Right now, I do not know how exactly it's generated, so due to lack of knowledge and overall experience in Xen (Xen policy)  I don't feel right writing the docs.

Hence, putting it on pause (or can close).

lorc:
>  one thing I know is that SSIDs can be found in xen/xen/xsm/flask/include/flask.h

I believe this is sufficient. Just mention somewhere in the readme that :

1. SSID support is very experimental and/or on early stages
2. SSIDs are generated during Xen compile time
3. Generated SSIDs can be found at xen/xen/xsm/flask/include/flask.h.

GrygiriiS:
> I believe this is sufficient. Just mention somewhere in the readme that :

readme updated

lorc (approved):
Thanks

`Reviewed-by: Volodymyr Babchul <volodymyr_babchuk@epam.com>`

firscity (approved):
Thank you

PR #96, opened 2024-07-19
xrun: update xrun_run to work with new behaviour of domain_create

oleksiimoisieiev:
On the latest version of the domain_create call it's work behavoiur was changed to return domid instead of 0 if operation was succeeded. Also, domain is started automatically so update xrun_run call to match the new device_create call revision.


firscity (approved):

oleksiimoisieiev:
@firscity updated

PR #97, opened 2024-07-22
xrun: storage: fix xrun_file_read_debounce() to return read bytes

GrygiriiS:
This PR fixes:

- xrun: storage: fix xrun_file_read_debounce() to return read bytes
- CONFIG_XRUN_STORAGE_DMA_DEBOUNCE default value to be set only for RPI5.

### DIFF
--- a/xrun/Kconfig
+++ b/xrun/Kconfig
@@ -47,7 +47,7 @@ config XRUN_IRQS_MAX
 
 config XRUN_STORAGE_DMA_DEBOUNCE
 	int "Set debounce buffer for FS storage access in KB"
-	default 4
+	default 0
 	help
 	  Sets debounce buffer for FS storage access in KB which is required to
 	  enable DMA for storage devices. The xenlib may use DMA incompatible
--- a/xrun/src/storage.c
+++ b/xrun/src/storage.c
@@ -19,11 +19,11 @@ static uint8_t debounce_buf[KB(CONFIG_XRUN_STORAGE_DMA_DEBOUNCE)]
 			    __aligned(CONFIG_SDHC_BUFFER_ALIGNMENT) __nocache;
 static K_MUTEX_DEFINE(debounce_lock);
 
-static int xrun_file_read_debounce(struct fs_file_t *file, uint8_t *buf, size_t read_size)
+static ssize_t xrun_file_read_debounce(struct fs_file_t *file, uint8_t *buf, size_t read_size)
 {
 	ssize_t read;
 	size_t count;
-	int ret = 0;
+	ssize_t ret = 0;
 
 	k_mutex_lock(&debounce_lock, K_FOREVER);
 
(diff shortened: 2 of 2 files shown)


xrun/src/storage.c
```diff
@@ -43,10 +43,14 @@ static int xrun_file_read_debounce(struct fs_file_t *file, uint8_t *buf, size_t
 		LOG_DBG("file count %zd read %zd", count, read);
 		count -= read;
 		buf += read;
+		if (count && read < sizeof(debounce_buf)) {
```
lorc, line 46: You can just check for `read==0` as this is the indication that `read_fs` reached end of file. At least this stands true for POSIX `read()` function ...

xrun/Kconfig
```diff
@@ -47,7 +47,8 @@ config XRUN_IRQS_MAX
 
 config XRUN_STORAGE_DMA_DEBOUNCE
 	int "Set debounce buffer for FS storage access in KB"
-	default 4
+	default 4 if BOARD_RPI_5
```
lorc, line 50: We don't want zephyr-xenlib to know about specific board(s). This is dependency in wrong direction.  Please remove this change.

xrun/src/storage.c
GrygiriiS: Not sure what you mean - Zephyr doc says https://docs.zephyrproject.org/apidoc/latest/group__file__system__api.html#gaba7e07127777187eedcd6976d352ab76

`A returned value may be lower than size if there were fewer bytes available than requested.`

And it sounds logically correct - file size is 5K. buf size is 4K. Read request is 8K (for example)
-  first read gets 4K.
-  second read 1K. But not 0.
- total 5K

xrun/Kconfig
GrygiriiS: Ok. But here to options possible:

1. Set it to 0 by default. and fix dom0-xt for RPI5
2. Keep as is - but rcar don't need it

?

lorc: Option 1, as it will not break existing setups.
Also, this feature is only required on boards with not IO-MMU, which are rare. So it should be disabled in most cases.

xrun/src/storage.c
lorc: Yeah, I was looking at POSIX docs, which add more context:

```
        On success, the number of bytes read is returned (zero indicates end of file), and the file position is advanced by this number.  It is not an error if this number is smaller than the
       number of bytes requested; this may happen for example because fewer bytes are actually available right now (maybe because we were close to end-of-file, or because we are reading from
       a pipe, or from a terminal), or because read() was interrupted by a signal. 
```

and

```
 If the file offset is at or past the
 end of file, no bytes are read, and read() returns zero.
```

Honestly, I don't know how much this is applicable to Zephyr, but I expect that they plan basically the behaviour: `ret = 0` indicates EOF and `ret > 0` indicates that there still be might more data.

GrygiriiS: Actually it's working as described above:

- file_pos is close to EOF. ret = EOF - file_pos
- file_pos is at EOF. ret = 0

xrun/Kconfig
GrygiriiS: fixed

firscity (approved):

al1img:
`Tested-by: Oleksandr Grytsov <oleksandr_grytsov@epam.com>`

lorc (approved):

PR #101, opened 2025-08-19
vch: use correct structure for grant unmapping

firscity:
Zephyr grant table driver used incorrect structure for grant entries unmapping. Recently it was fixed and now causes warnings during build of vchan related code, since it was developed according to previous functions signatures.

Rework vchan to use correct 'struct gnttab_unmap_grant_ref' for grant entries unmapping procedures. Since this structure uses grant handles instead of gref, add such member it to vch_handle struct.


sa-kib (approved):
`Acked-by: Sergiy Kibrik <sergiy_kibrik@epam.com>`

PR #103, opened 2025-10-13
xenstore: Add XenStore client library

### LINKED ISSUE #78: Add xenstore commandline interface for DomUs
Right now, `read`, `write` and `ls` commands are implemented as a part of xenstore server, so only Dom0 can use them. But it would be really nice if DomU users will be able to use them too.

soburi:
Introduce `xenstore-cli` module providing XenStore client function.

Related to this, the code that can be shared with xenstore-srv has been standardized as xenstore-common.

Note: 
I haven't yet been able to thoroughly test the error routes such as requiring server-side error injection.
I'd appreciate any tips on testing and CI.

Fix #78

### DIFF
--- a/xenstore-cli/include/xenstore_cli.h
+++ b/xenstore-cli/include/xenstore_cli.h
@@ -1,0 +1,366 @@
+/*
+ * Copyright (c) 2025 TOKITA Hiroshi
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef XENLIB_XENSTORE_CLI_H_
+#define XENLIB_XENSTORE_CLI_H_
+
+#include <sys/types.h>
+#include <stdint.h>
+#include <string.h>
+
+#include <xen/public/io/xs_wire.h>
+#include <xen/public/memory.h>
+#include <xen/public/xen.h>
+
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/spinlock.h>
+#include <zephyr/sys/slist.h>
+#include <zephyr/sys/device_mmio.h>
+#include <zephyr/xen/events.h>
+#include <zephyr/xen/generic.h>
+#include <zephyr/xen/hvm.h>
+
+#include <xenstore_common.h>
+
+/** Convenience constant for requests issued outside a XenStore transaction. */
+#define XS_TRANSACTION_NONE 0U
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief Watch notification callback.
+ *
+ * Invoked when a XenStore watch fires.
+ *
+ * @param path   Absolute XenStore path that triggered the watch.
+ * @param token  User-supplied token associated with the watch registration.
+ * @param param  Opaque user pointer provided at watcher initialization.
+ */
+typedef void (*xs_watch_cb)(const char *path, const char *token, void *param);
+
+/**
+ * @brief XenStore watcher descriptor.
+ *
+ * Initialize with xs_watcher_init() and register with ::xs_watcher_register().
+ * A watcher can be registered at most once at a time. Callbacks are invoked
+ * from the Xen event context; keep handlers short and non-blocking.
+ */
+struct xs_watcher {
+	sys_snode_t node;
+	xs_watch_cb cb;
+	void *param;
+};
+
+/**
+ * @brief Initialize the XenStore client.
+ * Safe to call multiple times.
+ *
+ * @retval 0       Initialization succeeded or was already active.
+ * @retval -errno  Initialization failed.
+ */
+int xs_init(void);
+
+/**
+ * @brief Set the default timeout for XenStore operations.
+ *
+ * Operations that do not take an explicit timeout parameter will use this value.
+ * K_FOREVER is the default, causing the calling thread to block indefinitely.
+ *
+ * @param tout  Default timeout to use for XenStore operations.
+ */
+void xs_set_default_timeout(k_timeout_t tout);
+
+/**
+ * @brief Prepare a watcher.
+ *
+ * @param w     Pointer of watcher descriptor to initialize.
+ * @param cb    Callback function invoked for watch notifications.
+ * @param param Opaque user data passed to the callback.
+ */
+void xs_watcher_init(struct xs_watcher *w, xs_watch_cb cb, void *param);
+
+/**
+ * @brief Register a watcher callback.
+ *
+ * @param w Watcher descriptor.
+ *
+ * @retval 0      Succeed to register the watcher.
+ * @retval -errno Failed to register.
+ */
+int xs_watcher_register(struct xs_watcher *w);
+
+/**
+ * @brief Read a XenStore path.
+ *
+ * @param path     The path to read the value.
+ * @param buf      A string buffer to store the result.
+ * @param len      The length of @p buf in bytes.
+ * @param tx_id    XenStore transaction identifier (0 when not in a transaction).
+ * @param tout     Timeout to wait for the response; use K_FOREVER to block
+ *                 indefinitely.
+ *
+ * @retval >=0     Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno  Request failed.
+ */
+ssize_t xs_read_timeout(const char *path, char *buf, size_t len, uint32_t tx_id, k_timeout_t tout);
+
+/**
+ * @brief Remove a XenStore entry.
+ *
+ * @param path     The path to entry to remove.
+ * @param buf      A string buffer to store the result.
+ * @param len      The length of @p buf in bytes.
+ * @param tx_id    XenStore transaction identifier (0 when not in a transaction).
+ *
+ * @retval >=0     Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno  Request failed.
+ *
+ * @see xs_rm_timeout().
+ */
+ssize_t xs_rm_timeout(const char *path, char *buf, size_t len, uint32_t tx_id, k_timeout_t tout);
+
+/**
+ * @brief Write a value to XenStore entry.
+ *
+ * @param path       The path to write the value.
+ * @param value      The value to write.
+ * @param value_len  The bytes of @p value.
+ * @param buf        A string buffer to store the result.
+ * @param len        The length of @p buf in bytes.
+ * @param tx_id      XenStore transaction identifier (0 when not in a transaction).
+ * @param tout       Timeout to wait for the response.
+ *
+ * @retval >=0       Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno    Request failed.
+ *
+ * @see xs_write_timeout().
+ */
+ssize_t xs_write_timeout(const char *path, const char *value, size_t value_len, char *buf,
+			 size_t len, uint32_t tx_id, k_timeout_t tout);
+
+/**
+ * @brief Set permission of XenStore entry.
+ *
+ * @param path       The path to set permission.
+ * @param perms      The list of permission specifier.
+ * @param perms_num  The number of @p perms entries.
+ * @param buf        A string buffer to store the result.
+ * @param len        The length of @p buf in bytes.
+ * @param tx_id      XenStore transaction identifier (0 when not in a transaction).
+ * @param tout       Timeout to wait for the response.
+ *
+ * @retval >=0       Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno    Request failed.
+ *
+ * @see xs_set_permissions_timeout().
+ */
+ssize_t xs_set_permissions_timeout(const char *path, const char **perms, size_t perms_num,
+				   char *buf, size_t len, uint32_t tx_id, k_timeout_t tout);
+
+/**
+ * @brief Get permission of XenStore entry.
+ *
+ * @param path     The path to get permission.
+ * @param buf      A string buffer to store the result.
+ * @param len      The length of @p buf in bytes.
+ * @param tx_id    XenStore transaction identifier (0 when not in a transaction).
+ * @param tout     Timeout to wait for the response.
+ *
+ * @retval >=0     Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno  Request failed.
+ *
+ * @see xs_get_permissions_timeout().
+ */
+ssize_t xs_get_permissions_timeout(const char *path, char *buf, size_t len, uint32_t tx_id,
+				   k_timeout_t tout);
+
+/**
+ * @brief Enumerate a XenStore directory.
+ *
+ * @param path     The absolute directory path to enumerate.
+ * @param buf      A buffer to store the result.
+ *                 The directory-entries are returned as a byte sequence of
+ *                 NUL-separated strings.
+ * @param len      The length of @p buf in bytes.
+ * @param tx_id    XenStore transaction identifier (0 when not in a transaction).
+ * @param tout     Timeout to wait for the response.
+ *
+ * @retval >=0     Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno  Request failed.
+ */
+ssize_t xs_directory_timeout(const char *path, char *buf, size_t len, uint32_t tx_id,
+			     k_timeout_t tout);
+
+/**
+ * @brief Start to watch the XenStore value changes.
+ *
+ * @param path     The absolute path to watch.
+ * @param token    A user token to identify watch request.
+ * @param buf      A buffer to store the result.
+ * @param len      The length of @p buf in bytes.
+ * @param tx_id    XenStore transaction identifier (0 when not in a transaction).
+ * @param tout     Timeout to wait for the response.
+ *
+ * @retval >=0     Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno  Request failed.
+ */
+ssize_t xs_watch_timeout(const char *path, const char *token, char *buf, size_t len, uint32_t tx_id,
+			 k_timeout_t tout);
+
+/**
+ * @brief Stop to watch the XenStore value changes.
+ *
+ * @param path     Absolute path to unwatch.
+ * @param token    A user token to identify watch request.
+ * @param buf      A string buffer to store the result.
+ * @param len      The length of @p buf in bytes.
+ * @param tx_id    XenStore transaction identifier (0 when not in a transaction).
+ * @param tout     Timeout to wait for the response.
+ *
+ * @retval >=0     Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno  Request failed.
+ *
+ * @see xs_watch_timeout().
+ */
+ssize_t xs_unwatch_timeout(const char *path, const char *token, char *buf, size_t len,
+			   uint32_t tx_id, k_timeout_t tout);
+
+/**
+ * @brief Read a XenStore path with the default timeout.
+ *
+ * @param path     The path to read the value.
+ * @param buf      A string buffer to store the result.
+ * @param len      The length of @p buf in bytes.
+ * @param tx_id    XenStore transaction identifier (0 when not in a transaction).
+ *
+ * @retval >=0     Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno  Request failed.
+ *
+ * @see xs_read_timeout().
+ */
+ssize_t xs_read(const char *path, char *buf, size_t len, uint32_t tx_id);
+
+/**
+ * @brief Remove a XenStore entry with the default timeout.
+ *
+ * @param path     The path to entry to remove.
+ * @param buf      A string buffer to store the result.
+ * @param len      The length of @p buf in bytes.
+ * @param tx_id    XenStore transaction identifier (0 when not in a transaction).
+ *
+ * @retval >=0     Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno  Request failed.
+ *
+ * @see xs_rm_timeout().
+ */
+ssize_t xs_rm(const char *path, char *buf, size_t len, uint32_t tx_id);
+
+/**
+ * @brief Write a value to XenStore entry with the default timeout.
+ *
+ * @param path       The path to write the value.
+ * @param value      The value to write.
+ * @param value_len  The bytes of @p value.
+ * @param buf        A string buffer to store the result.
+ * @param len        The length of @p buf in bytes.
+ * @param tx_id      XenStore transaction identifier (0 when not in a transaction).
+ *
+ * @retval >=0       Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno    Request failed.
+ *
+ * @see xs_write_timeout().
+ */
+ssize_t xs_write(const char *path, const char *value, size_t value_len, char *buf, size_t len,
+		 uint32_t tx_id);
+
+/**
+ * @brief Set permission of XenStore entry with the default timeout.
+ *
+ * @param path       The path to set permission.
+ * @param perms      The list of permission specifier.
+ * @param perms_num  The number of @p perms entries.
+ * @param buf        A string buffer to store the result.
+ * @param len        The length of @p buf in bytes.
+ * @param tx_id      XenStore transaction identifier (0 when not in a transaction).
+ *
+ * @retval >=0       Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno    Request failed.
+ *
+ * @see xs_set_permissions_timeout().
+ */
+ssize_t xs_set_permissions(const char *path, const char **perms, size_t perms_num, char *buf,
+			   size_t len, uint32_t tx_id);
+
+/**
+ * @brief Get permission of XenStore entry with the default timeout.
+ *
+ * @param path       The path to get permission.
+ * @param buf        A string buffer to store the result.
+ * @param len        The length of @p buf in bytes.
+ * @param tx_id      XenStore transaction identifier (0 when not in a transaction).
+ *
+ * @retval >=0       Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno    Request failed.
+ *
+ * @see xs_get_permissions_timeout().
+ */
+ssize_t xs_get_permissions(const char *path, char *buf, size_t len, uint32_t tx_id);
+
+/**
+ * @brief Enumerate a XenStore directory with the default timeout.
+ *
+ * @param path     The directory path to enumerate.
+ * @param buf      A string buffer to store the result.
+ * @param len      The length of @p buf in bytes.
+ * @param tx_id    XenStore transaction identifier (0 when not in a transaction).
+ *
+ * @retval >=0     Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno  Request failed.
+ *
+ * @see xs_directory_timeout().
+ */
+ssize_t xs_directory(const char *path, char *buf, size_t len, uint32_t tx_id);
+
+/**
+ * @brief Start to watch the XenStore value changes with the default timeout.
+ *
+ * @param path     Absolute path to watch.
+ * @param token    A user token to identify watch request.
+ * @param buf      A string buffer to store the result.
+ * @param len      The length of @p buf in bytes.
+ * @param tx_id    XenStore transaction identifier (0 when not in a transaction).
+ *
+ * @retval >=0     Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno  Request failed.
+ *
+ * @see xs_watch_timeout().
+ */
+ssize_t xs_watch(const char *path, const char *token, char *buf, size_t len, uint32_t tx_id);
+
+/**
+ * @brief Stop to watch the XenStore value changes with the default timeout.
+ *
+ * @param path     Absolute path to unwatch.
+ * @param token    A user token to identify watch request.
+ * @param buf      A string buffer to store the result.
+ * @param len      The length of @p buf in bytes.
+ * @param tx_id    XenStore transaction identifier (0 when not in a transaction).
+ *
+ * @retval >=0     Number of bytes copied into @p buf (terminator excluded).
+ * @retval -errno  Request failed.
+ *
+ * @see xs_watch_timeout().
+ */
+ssize_t xs_unwatch(const char *path, const char *token, char *buf, size_t len, uint32_t tx_id);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
--- a/xenstore-cli/src/xenstore_cli.c
+++ b/xenstore-cli/src/xenstore_cli.c
@@ -1,0 +1,955 @@
+/*
+ * Copyright (c) 2025 TOKITA Hiroshi
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <string.h>
+
+#include <xen/public/io/xs_wire.h>
+#include <xen/public/memory.h>
+#include <xen/public/xen.h>
+#include <xenstore_common.h>
+#include <xenstore_cli.h>
+
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/spinlock.h>
+#include <zephyr/sys/device_mmio.h>
+#include <zephyr/sys/slist.h>
+#include <zephyr/sys/util.h>
+#include <zephyr/sys/barrier.h>
+#include <zephyr/sys/atomic.h>
+
+#include <zephyr/xen/events.h>
+#include <zephyr/xen/generic.h>
+#include <zephyr/xen/hvm.h>
+
+LOG_MODULE_REGISTER(xenstore_cli);
+
+#define SZ_SOCKMSG  sizeof(struct xsd_sockmsg)
+#define SZ_FRAME(h) (SZ_SOCKMSG + h->len)
+
+/* Per-call waiter: caller owns the semaphore, worker side just signals it. */
+struct xenstore_response {
+	sys_snode_t node;
+	uint8_t *buf;
+	size_t len;
+	size_t pos;
+	struct k_sem sem;
+	uint32_t req_id;
+	int err;
+};
+
+/*
+ * Central client state shared by all APIs. Ring access and frame parsing are
+ * funneled through the dedicated work queue thread, so most members only need
+ * to be touched from a single context and can stay lock-free.
+ */
+struct xenstore_client {
+	struct xenstore_domain_interface *domint;
+	evtchn_port_t local_evtchn;
+
+	k_timeout_t default_timeout;
+	atomic_t next_req_id;
+	size_t to_discard;
+
+	struct k_spinlock lock;
+	struct k_mutex req_mutex;
+
+	/** Headers of the currently processed payload */
+	uint8_t hdr_buf[SZ_SOCKMSG];
+	size_t hdr_pos;
+
+	uint8_t work_buf[XENSTORE_PAYLOAD_MAX + 1];
+	struct xenstore_response work_resp;
+
+	/* resp_list: pending synchronous requests waiting on the worker to signal. */
+	sys_slist_t resp_list;
+	/* notify_list: registered watchers fired from the same frame pipeline. */
+	sys_slist_t notify_list;
+
+	K_KERNEL_STACK_MEMBER(workq_stack, CONFIG_XEN_STORE_CLI_WORKQ_STACK_SIZE);
+	struct k_work event_work;
+	struct k_work_q workq;
+	int workq_priority;
+	bool workq_started;
+};
+
+static struct xenstore_client xs_cli;
+
+/* Frame helpers keep the normal/exceptional paths from duplicating resets. */
+static void xs_frame_reset(struct xenstore_client *xs)
+{
+	xs->hdr_pos = 0;
+	xs->to_discard = 0;
+	xs->work_resp.pos = 0;
+	xs->work_resp.err = 0;
+	xs->work_resp.len = 0;
+}
+
+/* Lazily size the scratch buffer so the owning waiter receives the right span. */
+static void xs_frame_prepare_buffer(struct xenstore_client *xs, size_t capacity, uint32_t req_id)
+{
+	xs->work_resp.len = capacity;
+	xs->work_resp.pos = 0;
+	xs->work_resp.req_id = req_id;
+
+	xs->work_resp.node.next = NULL;
+	xs->work_resp.buf = xs->work_buf;
+	xs->work_resp.err = 0;
+
+	if (capacity > 0) {
+		memset(xs->work_resp.buf, 0, capacity);
+	}
+}
+
+static void xs_frame_mark_discard(struct xenstore_client *xs, size_t hdr_len)
+{
+	xs->hdr_pos = 0;
+	/* Remember the unread tail so the worker drains it before parsing again. */
+	xs->to_discard = hdr_len;
+	xs->work_resp.pos = 0;
+	xs->work_resp.err = 0;
+	xs->work_resp.len = 0;
+}
+
+static inline bool xs_is_initialized(const struct xenstore_client *xs)
+{
+	return xs && (xs->domint != NULL);
+}
+
+static inline uint32_t alloc_req_id(void)
+{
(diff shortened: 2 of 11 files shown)


soburi:
@klogg 

> @soburi mention it timplements #78

Added xenstore client commands. Thank you for your comment.

soburi:
@firscity @klogg @lorc

Could you review if you can take a time.

xenstore-cli/include/xenstore_cli.h
```diff
@@ -0,0 +1,368 @@
+/*
```
svlad-90, line 1: Hi @soburi, this header declares APIs returning ssize_t, but it does not include a header that defines ssize_t. A clean Zephyr module build failed in this repository before runtime validation could start.

```/home/builder/new_dev/zephyr-xenstore-client/dev/zephyr-xenlib-pr103-review/xenstore-cli/include/xenstore_cli.h:113:1: error: unknown type name 'ssize_t'; did you mean '_ssize_t'?```

Proposed fixes ( combined ): [pr103-xenstore-cli-review_combined_fixes.patch](https://github.com/user-attachments/files/30344989/pr103-xenstore-cli-review_combined_fixes.patch)

File: xenstore-cli/src/xenstore_cli.c [hunk 2]
```diff
@@ -0,0 +1,1112 @@
...
+
+	uint8_t work_buf[XENSTORE_PAYLOAD_MAX + 1];
+	struct xenstore_response work_resp;
+
+	/* resp_list: pending synchronous requests waiting on the worker to signal. */
+	sys_slist_t resp_list;
+	/* notify_list: registered watchers fired from the same frame pipeline. */
+	sys_slist_t notify_list;
+
+	K_KERNEL_STACK_MEMBER(workq_stack, CONFIG_XEN_STORE_CLI_WORKQ_STACK_SIZE);
```
svlad-90: Hi @sobur, please define the private workqueue stack as a top-level K_KERNEL_STACK_DEFINE()
  / K_THREAD_STACK_DEFINE() object instead of embedding it in struct
  xenstore_client with K_KERNEL_STACK_MEMBER. The member macro only provides
  embedded stack storage; it does not make the stack a normal top-level Zephyr
  kernel stack object. On ARM64 Zephyr/Xen this is trapped on the first response
  event, while entering Zephyr's exception-vector register save path. A top-level
  kernel stack has the placement/alignment expected by k_work_q's backing thread.

Image

Proposed fixes ( combined ): [pr103-xenstore-cli-review_combined_fixes.patch](https://github.com/user-attachments/files/30344989/pr103-xenstore-cli-review_combined_fixes.patch)

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1112 @@
...
+
+	/* Single-threaded pump: consume new bytes, parse at most one frame per iteration. */
+	while (true) {
+		ret = ring_avail_for_read(xs, &avail);
+		if (ret < 0) {
+			fail_pending_responses(xs, ret);
+			break;
+		}
+
+		if (avail == 0) {
```
svlad-90, line 1072: Hi @soburi, please do not rely on the event-channel callback as the only RX trigger while synchronous requests are pending. The local Zephyr/Xen QEMU runtime can miss that wakeup path; the request is sent, but the caller waits forever. Track whether requests are pending and let RX work briefly poll/resubmit until a response appears or the caller's timeout handles the operation.

Image

Without fix, tests are stuck:

Image

With the fix, it is OK:

Image

Proposed fixes ( combined ): [pr103-xenstore-cli-review_combined_fixes.patch](https://github.com/user-attachments/files/30344989/pr103-xenstore-cli-review_combined_fixes.patch)

xenstore-cli/include/xenstore_cli.h
```diff
@@ -0,0 +1,368 @@
...
+ * @param path   Absolute XenStore path that triggered the watch. The pointer is
+ *               only valid for the duration of the callback.
+ * @param token  User-supplied token associated with the watch registration. The
+ *               pointer is only valid for the duration of the callback.
+ * @param param  Opaque user pointer provided at watcher initialization.
+ */
+typedef void (*xs_watch_cb)(const char *path, const char *token, void *param);
+
+/**
+ * @brief XenStore watcher descriptor.
```
svlad-90, line 66: Hi @soburi, please add a watcher unregister API and define callback lifetime semantics. If a callback is already running, unregister should either wait for it to finish or document a safe non-blocking ownership contract.

PR #117 now has the descriptor model and unregister semantics pushed; see:

- https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/include/xenstore_client.h#L56
- https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/include/xenstore_client.h#L312
- https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/include/xenstore_client.h#L333.

xenstore-cli/include/xenstore_cli.h
```diff
@@ -0,0 +1,368 @@
...
+ * short and non-blocking.
+ */
+struct xs_watcher {
+	sys_snode_t node;
+	xs_watch_cb cb;
+	void *param;
+};
+
+/**
+ * @brief Initialize the XenStore client.
```
svlad-90, line 105: Hi @soburi, please serialize xs_init()/connect inside the library. One caller should perform setup, concurrent callers should wait for the result, and partial failures should leave a clearly reusable or failed state. PR #117 does this in the connect path; see https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/src/xenstore_client.c#L210.

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1112 @@
...
+			(uint32_t)prod);
+		*avail = 0;
+		return -EIO;
+	}
+
+	*avail = XENSTORE_RING_SIZE - (prod - cons);
+	return 0;
+}
+
+static int ring_write_all(struct xenstore_client *xs, const void *buf, size_t len)
```
svlad-90: Hi @soburi, please make request-ring write timeout-aware. A full request ring should wait for the server to consume bytes until the operation timeout expires, while still detecting invalid ring indexes and waking the server. PR #117's xs_write_all() shows this pattern; see https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/src/xenstore_client.c#L341.

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1112 @@
...
+		xs->hdr_pos += ret;
+		*avail -= ret;
+
+		if (xs->hdr_pos < SZ_SOCKMSG) {
+			LOG_DBG("header not ready");
+			return -EAGAIN;
+		}
+	}
+
+	if (hdr->len > XENSTORE_PAYLOAD_MAX) {
```
svlad-90, line 664: Hi @soburi, please use one fatal transport path for ring-index corruption, impossible headers, oversized frames, and local ring read/write failures. That path should mark the client disconnected/failed, wake all pending callers, and make future calls fail until reconnect. PR #117 has xs_mark_transport_failed() and xs_fail_all_pending(); see:

- https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/src/xenstore_client.c#L168
- https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/src/xenstore_client.c#L153.

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1112 @@
...
+		LOG_ERR("payload too large: %u > " STRINGIFY(XENSTORE_PAYLOAD_MAX), hdr->len);
+		fail_client(xs, -EMSGSIZE);
+		return -EMSGSIZE;
+	} else if ((hdr->type != XS_WATCH_EVENT) && (hdr->req_id == 0)) {
+		LOG_ERR("Invalid response header: req_id must be non-zero");
+		ret = -EPROTO;
+	} else if ((hdr->type == XS_WATCH_EVENT) && (hdr->req_id != 0)) {
+		LOG_ERR("Invalid watch header: req_id=%u (expected 0)", hdr->req_id);
+		ret = -EPROTO;
+	} else if (hdr->len > xs->work_resp.len) {
```
svlad-90: Hi @soburi, please separate valid-but-too-large replies from corrupt protocol frames. A small user buffer should complete that request with -EMSGSIZE and drain exactly that reply, without poisoning the client; invalid wire framing should use the fatal transport path. PR #117 keeps a receive scratch buffer independent from the caller output buffer and completes the pending request with the right errno.

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1112 @@
...
+		err = xenstore_get_error(xs->work_resp.buf,
+					 MIN(xs->work_resp.pos, sizeof(xs->work_buf)));
+		if (err == 0) {
+			err = -EINVAL;
+		} else {
+			err = -err;
+		}
+	}
+
+	pending = lock_response(xs, xs->work_resp.req_id, &key);
```
svlad-90: Hi @soburi, please tighten the pending-request lifecycle so one side owns completion and unlinking. That reduces timeout/reply races and makes fail-all-pending behavior simpler. PR #117's xs_complete_request() removes the request under the pending lock before waking the waiter; see https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/src/xenstore_client.c#L471.`

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1112 @@
...
+	size_t avail;
+	int ret;
+
+	if (!xs_is_initialized(xs)) {
+		return;
+	}
+
+	/* Single-threaded pump: consume new bytes, parse at most one frame per iteration. */
+	while (true) {
+		ret = ring_avail_for_read(xs, &avail);
```
svlad-90: Hi @soburi, please treat response-ring index corruption as a fatal transport error. Wake all pending callers and transition the client to a failed/disconnected state so future calls fail cleanly until reconnect. PR #117 uses xs_mark_transport_failed() for this class of error; see https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/src/xenstore_client.c#L168 and https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/src/xenstore_client.c#L406.

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1112 @@
...
+		err = -ENODEV;
+		goto end;
+	}
+
+	err = ring_avail_for_write(xs, &avail);
+	if (err < 0) {
+		goto end;
+	}
+
+	if (avail < (SZ_SOCKMSG + hdr.len)) {
```
svlad-90: Hi @soburi, please make request-ring backpressure honor the caller's timeout. A full ring should not become an immediate EAGAIN after one yield for a synchronous API that accepts K_FOREVER or a long timeout. PR #117's timeout-aware xs_write_all() is the behavior I would expect here; see https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/src/xenstore_client.c#L341.`

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1112 @@
...
+	}
+
+	if (resp_local.pos > 0) {
+		result = MIN(len, resp_local.pos);
+	}
+
+	return result;
+}
+
+int xs_init(void)
```
svlad-90, line 1444: Hi @soburi, please split the connect state from the raw domint pointer and serialize initialization. A connect state machine should handle not-started, connecting, connected, and failed/teardown states so concurrent callers and partial failures are deterministic. PR #117's xs_client_connect() is the reference point; see https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/src/xenstore_client.c#L210.

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1112 @@
...
+	}
+	xs_cli.local_evtchn = value;
+
+	err = hvm_get_parameter(HVM_PARAM_STORE_PFN, DOMID_SELF, &paddr);
+	if (err) {
+		LOG_ERR("hvm_get_param(STORE_PFN) failed: err=%d", err);
+		return -EIO;
+	}
+
+	device_map(&vaddr, XEN_PFN_PHYS(paddr), XEN_PAGE_SIZE, K_MEM_CACHE_WB | K_MEM_PERM_RW);
```
svlad-90, line 1497: Hi @soburi, please validate STORE_EVTCHN and STORE_PFN, map the shared XenStore page with device/non-cacheable attributes, and handle XENSTORE_RECONNECT explicitly. PR #117 does this in the client connect path; see:

- https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/src/xenstore_client.c#L223
- https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/src/xenstore_client.c#L263
- https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xenstore-client/src/xenstore_client.c#L285.

The Dom0-side page allocation/reuse flow is described in https://github.com/svlad-90/zephyr-xenlib/blob/xt-xenstore-client-review/xen-dom-mgmt/src/xen-dom-mgmt.c#L1102.

svlad-90:
Hi @soburi,

I also worked on a XenStore client implementation in parallel  ( located here - https://github.com/xen-troops/zephyr-xenlib/pull/118 ). After looking at your PR, I saw that your changes cover a broader integration area and are more tightly connected to your current xenstore-cli direction, so I think it makes sense for your implementation to be the base one.

Instead of pushing a competing client implementation, I used my branch mainly as a validation/reference point and reviewed your implementation against the local Zephyr XenStore service. Some of my commits are not split as cleanly as a final reviewable series should be, but they capture the fixes and design patterns that passed local build/runtime validation.

I will reference that branch in the review comments where it helps explain a finding or show one possible implementation approach. Please treat those links as examples, not as a request to copy the code one-to-one.

svlad-90 (changes requested):

svlad-90:
Hi @soburi, I'm preparing fixes for the findings I raised, based on your implementation. I'm also adding a small PV frontend to validate the changes. Please give me a few days to finalize this.

soburi:
> Hi @soburi, I'm preparing fixes for the findings I raised, based on your implementation. I'm also adding a small PV frontend to validate the changes. Please give me a few days to finalize this.

@svlad-90 

Thank you. There's no need to rush, but I'm very happy to hear from you.

svlad-90:
Hi @soburi, thanks for waiting! I'm finally ready. ))

Please review the changes and let me know if I should make any further adjustments.

P.S. I've made more than just minor fixes. But I've tried to check it properly, including automated stress tests and testing on both a single CPU and an SMP setup.

----

# XenStore Client Fixes Review Note

Target branch:

https://github.com/svlad-90/zephyr-xenlib/commits/xenstore-client-extension/

Current patch series:

```text
bea4b14 xen-blkfront: add minimal PV block frontend
c8750f2 xenstore-cli: fix client API and runtime behavior ( my adjustments )
106b07f xenstore: Add XenStore client library ( original commit )
```

Full runtime log:

[qemu-xen421-zephyr-vpl011-e2e-60s.log](https://github.com/user-attachments/files/30636515/qemu-xen421-zephyr-vpl011-e2e-60s.log)

## Summary

The client fix commit is larger than a single race fix. It hardens the public
API boundary, client lifecycle, timeout behavior, response worker progress,
watcher unregister semantics, typed permission handling, and the SMP timeout
path where submitted requests can outlive their caller waiters.

The blkfront commit is intentionally minimal. It is not meant to prove a full
Zephyr block driver yet. Its value is as a realistic XenStore consumer: it
drives a XenBus frontend lifecycle against a Linux blkback backend, publishes
ring and event-channel state through XenStore, reads backend metadata, performs
real shared-ring reads, and closes through a bounded XenBus shutdown path.

## Client Changes

- Make `xenstore_cli.h` self-contained by including the public `ssize_t`
  dependency directly. Users can include the client header without relying on
  include order from unrelated headers.

- Keep `xs_wire.h` private to `xenstore_cli.c`. The public API no longer
  exposes XenStore wire-protocol internals just to declare client functions.

- Use a top-level `K_KERNEL_STACK_DEFINE` object for the private workqueue
  stack. Some Zephyr architectures require kernel stacks to live in special
  sections. Embedding it in client state broke the local ARM64 Xen run.

- Serialize `xs_init()` with an init mutex. Concurrent callers now see one
  transport setup path instead of racing shared ring mapping, event-channel
  binding, workqueue setup, and list initialization.

- Track client readiness separately from the mapped ring pointer. API calls
  test a published initialized flag, while request submission and teardown
  serialize access to the shared ring mapping.

- Document the lifecycle boundary. Repeated `xs_init()` calls are safe after a
  successful init, but the library still has no public shutdown API and does
  not promise post-failure recovery through another `xs_init()` call.

- Apply operation timeouts while waiting for request-ring space and while
  writing request bytes. Timeout-enabled APIs no longer block indefinitely
  behind a full request ring.

- Submit the response worker after request submission and keep it making
  progress while callers are pending. This avoids waiting only for a later
  event-channel interrupt after the backend has been notified.

- Wake all pending callers when the response stream hits a fatal transport or
  parser error. Waiters now get the failure instead of sleeping until timeout
  or forever.

- Track already-submitted requests whose callers time out locally as bounded
  tombstones. Late backend replies are matched by `req_id` and drained as
  expected timed-out responses instead of accumulating as stale frames under
  SMP load.

- Share the fixed 64 request lifecycle slots between pending waiters and
  timeout tombstones. New requests return `-EBUSY` while that capacity is
  saturated, so every admitted request can still move from pending to a
  tombstone if its caller times out.

- Unlink waiters on normal completion, submit failure, and timeout.
  Stack-owned waiter nodes no longer remain visible to the worker after the
  caller stops waiting.

- Synchronize default-timeout reads and writes. Each default-timeout wrapper
  takes one stable `k_timeout_t` snapshot without racing
  `xs_set_default_timeout()`.

- Define watcher descriptor and callback parameter lifetimes in the public
  header. Callers can see that both the descriptor and callback-owned storage
  must outlive registration.

- Add `xs_watcher_unregister()`. Registered callbacks can now be removed before
  their descriptors or callback-owned storage are released.

- Make watcher unregister safe for in-flight callbacks. Other threads wait for
  a running callback, while self-unregister from the callback only removes
  future dispatch and avoids deadlock.

- Re-read the watcher list after each callback dispatch. Removed watchers are
  no longer reached through a stale next pointer.

- Document the split between local watcher registration and remote XenStore
  `XS_WATCH`/`XS_UNWATCH` operations. Local watchers receive every delivered
  watch event and must filter by path and token.

- Replace raw permission strings in the public API with typed permission
  entries. Applications no longer need to hand-build values such as `"r1"` or
  `"b3"`; the client owns wire serialization and parsing.

- Make set-permissions status-only. The API returns 0 on success instead of
  exposing the backend acknowledgment through caller-provided `buf`/`len`.

- Bound get-permissions temporary allocation by caller capacity. Permission
  reads allocate only enough wire buffer for entries the caller can receive.

- Document the set-permissions entry limit and validation behavior. Empty
  lists, oversized lists, and invalid typed permissions now fail predictably.

- Trim unused implementation includes. The source now carries only the Xen,
  Zephyr, and C headers used by the implementation.

- Group private helpers by role, simplify response frame handling, and keep
  protocol failures explicit while preserving per-request error handling.

- Add focused comments for private state, frame parsing, polling, lock context,
  and static helper roles. The response pipeline is easier to review without
  reconstructing the XenStore ring state machine from scratch.

## Validation

Validation was done with a local Xen/Zephyr runtime probe rather than only with
unit-level reasoning. The probe boots a Zephyr DomU with SMP enabled against a
Linux service domain with `xenstored` and `blkback`, then runs the XenStore
client through normal operations, invalid API calls, watcher
register/unregister/dispatch, request/watch stress, a concurrent XenStore
stress phase, and the minimal blkfront consumer.

Validation was run on a Zephyr branch that includes SMP-safe Xen event
  delivery, specifically commit 817f25dd497 ("drivers: xen: fix event delivery
  on SMP guests" - https://github.com/xen-troops/zephyr/pull/142). Without that base support, Xen event-channel delivery on
  secondary vCPUs is not a reliable test environment for the client patch.

  The client changes were also tested in a single-vCPU DomU configuration. That
  run does not exercise cross-vCPU event delivery, but it is useful as a baseline
  that the API/lifecycle changes do not depend on SMP-only behavior.

The concurrent phase starts:

- one normal read/write/permission actor
- one invalid-contract actor
- one short-timeout missing-read actor
- one watcher actor
- four read/write/remove/list-permission workers

Relevant runtime log excerpts:

```text
[xen] (XEN) DOM1: PR103 REVIEW PROBE START
[xen] (XEN) DOM1: pr103_case: negative-contracts PASS
[xen] (XEN) DOM1: pr103_case: request-stress PASS
[xen] (XEN) DOM1: pr103_case: watch-stress PASS
[xen] (XEN) DOM1: pr103_case: concurrent-stress START
[xen] (XEN) DOM1: pr103_case: concurrent-stress cpu-pin SKIP threads=8 cpus=2
[xen] (XEN) DOM1: pr103_case: xenstore-concurrent-rw-rm-list-perms PASS
[xen] (XEN) DOM1: pr103_case: concurrent-stress PASS
```

The blkfront consumer also passed in the same run:

```text
[xen] (XEN) DOM1: pr103_case: pvblock-negative PASS
[xen] (XEN) DOM1: pr103_case: pvblock-range-check PASS
[xen] (XEN) DOM1: pr103_case: pvblock-read PASS
[xen] (XEN) DOM1: pr103_case: pvblock-read-stress PASS reads=65
[xen] (XEN) DOM1: pr103_case: pvblock-concurrent-read PASS threads=4 reads=64
[xen] (XEN) DOM1: pr103_case: pvblock-close PASS
[xen] (XEN) DOM1: pr103_case: pvblock-frontend PASS
```

Final runtime result:

```text
[xen] (XEN) DOM1: PR103 REVIEW PROBE PASS
[dom0] QEMU_SMOKE_RC=124
```

`QEMU_SMOKE_RC=124` is expected for this harness because QEMU is terminated by
the outer timeout after the success markers are already printed.

Here is the test implementation, in case you want to look at it:
[main.c](https://github.com/user-attachments/files/30636731/main.c)

xenstore-cli/include/xenstore_cli.h
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d.

Fixed by including <sys/types.h> in xenstore_cli.h.

xenstore-cli/src/xenstore_cli.c
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d.

Fixed by using top-level K_KERNEL_STACK_DEFINE(xs_workq_stack, ...).

xenstore-cli/src/xenstore_cli.c
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d

Fixed with response retry work and worker progress while callers are pending.

xenstore-cli/include/xenstore_cli.h
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d

Fixed with xs_watcher_unregister(), callback-running tracking, and documented lifetime rules.

xenstore-cli/include/xenstore_cli.h
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d

Fixed with xs_init_lock and safe repeated initialization.

xenstore-cli/src/xenstore_cli.c
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d

Fixed by making request-ring backpressure wait until the caller deadline expires.

xenstore-cli/src/xenstore_cli.c
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d

Fixed with fail_transport(), readiness teardown, and wakeup of pending callers.

xenstore-cli/src/xenstore_cli.c
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d

Fixed so caller-buffer capacity failures complete that request, while protocol corruption remains fatal.

xenstore-cli/src/xenstore_cli.c
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d

Fixed by tightening waiter unlink/completion, plus timeout tombstones for late replies.

xenstore-cli/src/xenstore_cli.c
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d

Fixed by routing response-ring index errors through the fatal transport path.

xenstore-cli/src/xenstore_cli.c
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d

Fixed by the timeout-aware request-ring backpressure.

xenstore-cli/src/xenstore_cli.c
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d

Mostly addressed: there is a separate initialized flag, xs_init_lock, and deterministic repeated init. It is not a full explicit enum state machine.

xenstore-cli/src/xenstore_cli.c
svlad-90: Addressed in the proposed patch https://github.com/svlad-90/zephyr-xenlib/commit/c8750f272efd2f52aca6eff5558753c73578f35d

Partially obsolete / non-blocking. The non-cacheable shared-ring mapping was fixed. The explicit STORE_* invalid-value checks and XENSTORE_RECONNECT handling were raised during earlier Zephyr Dom0 XenStore-server validation, but the final Linux xenstored validation does not require that.

soburi:
@svlad-90 

> Hi @soburi, thanks for waiting! I'm finally ready. ))
> 
> Please review the changes and let me know if I should make any further adjustments.

Thank you for your suggestion.
This is mostly applicable, so I incorporate your tree.
but I noticed some points.

https://github.com/svlad-90/zephyr-xenlib/blob/xenstore-client-extension/xenstore-cli/src/xenstore_cli.c#L1799

This is huge consumption of stack resources.
Here, we only need to reserve for SET_PERMS response.
IMO, we can reduce this size.


https://github.com/svlad-90/zephyr-xenlib/blob/xenstore-client-extension/xenstore-cli/src/xenstore_cli.c#L1499

I think this is better to use K_MEM_CACHE_WB like as already implemented gnttab driver.

svlad-90:
Hi @soburi, thanks for accepting my adjustments!

Regarding your findings. I agree with them.
The adjusted implementation still passes the automated test suite. I've checked that.

I'll ask the maintainers (my colleagues) whether any additional review is required by them.

soburi:
> I'll ask the maintainers (my colleagues) whether any additional review is required by them.

It looks like you've already been contacted, but I'm mentioning you here just in case. > @firscity @lorc

lorc:
This is quite substantial change and it requires time to review it in the fullness. I'll check it, but probably not in the nearest time.

klogg:
@soburi please rebase and push your latest changes, if anything left - I think reviewers will be able to check it next week

I really want to prioritize and finish this task quickly!

soburi:
> @soburi please rebase and push your latest changes, if anything left - I think reviewers will be able to check it next week
> 
> I really want to prioritize and finish this task quickly!

I updated that yesterday, so it's all set.

xenstore-cli/include/xenstore_cli.h
```diff
@@ -0,0 +1,462 @@
...
+#include <zephyr/kernel.h>
+#include <zephyr/sys/slist.h>
+
+/** Convenience constant for requests issued outside a XenStore transaction. */
+#define XS_TRANSACTION_NONE 0U
+
+/**
+ * @brief XenStore access rights for one domain permission entry.
+ */
+enum xs_permission {
```
lorc, line 23: We already have `enum xs_perm` in xss.h. No need for the second instance.

I believe you can factor out it into some common header file.

xenstore-cli/include/xenstore_cli.h
```diff
@@ -0,0 +1,462 @@
...
+ * @brief XenStore watcher descriptor.
+ *
+ * Initialize with xs_watcher_init() and register with ::xs_watcher_register().
+ * A watcher can be registered at most once at a time and must remain valid for
+ * as long as it is registered. Use xs_watcher_unregister() before releasing
+ * the descriptor or callback-owned storage.
+ * Callbacks are invoked from the xenstore-cli workqueue thread; keep handlers
+ * short and non-blocking.
+ *
+ * All fields are internal to the library. Applications may allocate this
```
lorc, line 75: I believe this is a good idea to split this header to two: public API and internal data.

xenstore-cli/include/xenstore_cli.h
```diff
@@ -0,0 +1,462 @@
...
+ * exact lifetime rules.
+ *
+ * The @p param pointer is stored as-is and passed to @p cb. If @p cb
+ * dereferences it, the pointed object must remain valid for as long as a
+ * callback can run. A normal xs_watcher_unregister() call from another thread
+ * provides that quiescence when it returns. If the watcher unregisters itself
+ * from its callback, the object must remain valid until the current callback
+ * returns to the library.
+ *
+ * @param w     Pointer to the watcher descriptor to initialize.
```
lorc, line 144: I think, the prefered way is to hide internal representation from the clients. In this case, the library should allocate watcher structure on its own and return an opaque pointer to the caller. 

Also, this will make lifetime control simpler.

xenstore-cli/include/xenstore_cli.h
```diff
@@ -0,0 +1,462 @@
...
+ * @param tx_id     XenStore transaction identifier (0 when not in a transaction).
+ *
+ * @retval 0        Succeeded in setting permissions.
+ * @retval -EINVAL  Invalid argument, including @p perms_num equal to zero.
+ * @retval -E2BIG   @p perms_num is greater than 32.
+ * @retval -errno   Request failed.
+ *
+ * @see xs_set_permissions_timeout().
+ */
+int xs_set_permissions(const char *path, const struct xs_permission_entry *perms, size_t perms_num,
```
lorc, line 216: xen store server provides similar call. I think we need to provide exact same interface for apps that run on xenstore-server domain and on xenstore-client domain.

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1847 @@
...
+#include <zephyr/sys/slist.h>
+#include <zephyr/sys/util.h>
+#include <zephyr/sys/barrier.h>
+#include <zephyr/sys/atomic.h>
+
+#include <zephyr/xen/events.h>
+#include <zephyr/xen/generic.h>
+#include <zephyr/xen/hvm.h>
+
+LOG_MODULE_REGISTER(xenstore_cli);
```
lorc, line 32: In my opinion, this name is misleading: CLI stands for "Command Line Interface" and this is what I expect when I see "cli". Maybe better call it "xenstore_client"?

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1847 @@
...
+ */
+
+/*
+ * Common helpers cover small conversions and state resets shared by
+ * more than one pipeline stage. Keeping them first avoids local forward
+ * declarations for client/request/response blocks below.
+ */
+
+/* Convert the public permission enum to XenStore's one-character wire form. */
+static int perm_to_wire(enum xs_permission perm, char *wire)
```
lorc, line 183: Again, we have similar function in xenstore_srv.c. I believe it is better to move this into common code.

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1847 @@
...
+	case 'b':
+		*perm = XS_PERMISSION_READ_WRITE;
+		return 0;
+	default:
+		return -EINVAL;
+	}
+}
+
+/* Parse a NUL-separated XS_GET_PERMS payload into typed permission entries. */
+static ssize_t perm_parse_wire(const char *raw, size_t raw_len, struct xs_permission_entry *perms,
```
lorc, line 229: Looks pretty much like `deserialize_perms` from xenstore_srv

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1847 @@
...
+		perms[copied].perm = perm;
+		copied++;
+		off += entry_len + 1;
+	}
+
+	return copied;
+}
+
+/* Reset the staged response frame state. */
+static void frame_reset(struct xenstore_client *xs)
```
lorc, line 284: I noticed that there is only one instance of xenstore_client is present, which is a global variable defined in this file. I don't think that you need to pass pointer to it here.

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1847 @@
...
+
+	return 0;
+}
+
+/* Select the pending waiter or watch buffer that will receive the payload. */
+static int select_frame_buffer(struct xenstore_client *xs)
+{
+	struct xsd_sockmsg *hdr = (struct xsd_sockmsg *)(xs->hdr_buf);
+
+	if (hdr->type != XS_WATCH_EVENT) {
```
lorc, line 678: Better to have `switch (hdr->type)` to be able to filter out unknown messages

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1847 @@
...
+
+	if (to_read == 0) {
+		return 0;
+	}
+
+	ret = response_ring_read(xs, xs->frame.buf + xs->frame.pos, to_read);
+	if (ret < 0) {
+		LOG_ERR("ring_read failed while fetching type=%u req_id=%u: %d", hdr->type,
+			hdr->req_id, ret);
+		*fatal = true;
```
lorc, line 763: is `fatal` really needed? As I can see, non-fatal error code is only `EAGAIN`. So caller can determine "fatality" of error just by checking the error code

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1847 @@
...
+
+	return 0;
+}
+
+/* Dispatch a completed watch-event frame; caller must be the response worker. */
+static void watch_event_dispatch(struct xenstore_client *xs)
+{
+	const size_t payload_len = xs->frame.pos;
+	const char *path = "";
+	const char *token = "";
```
lorc, line 845: Why these two pointer are being initialized with empty string instead of `NULL`?

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1847 @@
...
+	key = k_spin_lock(&xs->lock);
+	dispatch_gen = xs->watch_dispatch_gen + 1;
+	if (dispatch_gen == 0) {
+		sys_snode_t *node;
+
+		/* Keep zero as the freshly initialized marker after wraparound. */
+		SYS_SLIST_FOR_EACH_NODE(&xs->notify_list, node) {
+			struct xs_watcher *w = CONTAINER_OF(node, struct xs_watcher, node);
+
+			w->dispatch_gen = 0;
```
lorc, line 881: I am not sure I got the idea behind this action. Could you please clarify why do you need to zero-out generation after wraparound?

xenstore-cli/include/xenstore_cli.h
lorc: @svlad-90, could you please explain what problem fixes callback-running tracking? I am not sure I am following this

xenstore-common/src/xenstore_common.c
```diff
@@ -75,8 +75,19 @@ int xenstore_get_error(const char *errstr, size_t len)
 {
 	size_t i;
 
+	if (!errstr) {
```
lorc, line 78: Please move this change to separate commit

xenstore-common/src/xenstore_common.c
```diff
@@ -75,8 +75,19 @@ int xenstore_get_error(const char *errstr, size_t len)
...
+		return 0;
+	}
+
 	for (i = 0; i < ARRAY_SIZE(xsd_errors); i++) {
-		if (strncmp(errstr, xsd_errors[i].errstring, len) == 0) {
+		const char *known = xsd_errors[i].errstring;
```
lorc, line 83: Could you please explain this change? Also, I believe it should go into a separate commit.

xenstore-common/include/xenstore_common.h
```diff
@@ -139,7 +139,8 @@ int xenstore_ring_write(struct xenstore_domain_interface *intf, const void *data
  * @brief Read data from the XenStore ring buffer.
  *
  * @param intf   Domain interface describing the shared ring.
- * @param data   Destination buffer that receives the data.
+ * @param data   Destination buffer that receives the data, or NULL to discard
```
lorc, line 142: That should be separate change

xenstore-cli/src/xenstore_cli.c
```diff
@@ -0,0 +1,1847 @@
...
+		w->callback_running = true;
+		k_spin_unlock(&xs->lock, key);
+
+		if (w->cb) {
+			w->cb(path, token, w->param);
+		}
+
+		key = k_spin_lock(&xs->lock);
+		w->callback_running = false;
+		k_sem_give(&w->callback_idle);
```
lorc, line 917: There is lots of logic to track if callback is running. What is the purpose of it?

svlad-90:
Hi @soburi, I'll take care of passing the first iteration of the review. I'll create a separate branch with all the fixes from this PR and post it here so you can review it against @lorc's comments and take it over if everything looks good.

Let me know if it is OK for me to proceed. ))

soburi:
> Hi @soburi, I'll take care of passing the first iteration of the review. I'll create a separate branch with all the fixes from this PR and post it here so you can review it against @lorc's comments and take it over if everything looks good.
> 
> Let me know if it is OK for me to proceed. ))

Understood. Please respond to the initial review. 
I think it's also necessary to reflect it in the PR. Please let me know.

svlad-90:
Hi @soburi, FYI.

The following points on the server side were addressed in the following PR, which has already been merged:
https://github.com/xen-troops/zephyr-xenlib/pull/132

- Factor out common logic with Xenstore-server and move it into common code. In a separate patch.
- Harmonize API with server's API, so the same code can work with server and client

My next step is to implement a watcher mechanism on the server side, and then move the client implementation on top of the common API, considering other @lorc's findings.

JarmouniA (changes requested):
Usage of Zephyr APIs in HAL repo shall be fixed first.
https://github.com/xen-troops/zephyr-xenlib/pull/132/files#r4088525240

PR #104, opened 2025-10-30
Integrate `zephyr-xenlib` to Zephyr's main repository

soburi:
# Summary

Proposing adoption of this project as a standard Zephyr module, to be merged into the main Zephyr repository.
Once approved and merged, I plan to apply for the repository to be added to Zephyr.

This change mainly imports Xen’s public headers.

Related Zephyr side work is tracked in https://github.com/zephyrproject-rtos/zephyr/pull/91643.

# Details

Zephyr recently tightened its policies against including non-Apache-2.0 licensed files in the main tree, as documented at https://github.com/zephyrproject-rtos/zephyr/tree/main/LICENSES

Adding any new license now requires TSC approval and is enforced by CI. Adding the MIT-licensed Xen public headers directly into the main repository is becoming cumbersome. Modules outside the main tree are exempt from this restriction, so the headers need to live in an external module instead.

The `zephyr-xenlib` module already brings in Xen public headers separately for the `vch` and `xenstore-srv` components. This proposal generalizes that approach so the module itself carries the Xen public headers, and it updates include paths accordingly.

# Verification

Functionally, this is purely a relocation of headers: no APIs or implementations change. However, when importing the Xen public headers, I applied local adjustments; please focus review attention there. (https://github.com/xen-troops/zephyr-xenlib/pull/102/commits/ea9b3eb71e0a35cbc8bd2d0ada2d5a58db4825e0)

I've also reapplied the commit history in the Zephyr repository to ensure a clear history. However, since it eventually replaces the content with that of xen-4.17, I'll squash if it was redundant.

The module has been validated in the https://github.com/xen-troops/meta-xt-prod-devel-rpi5 environment, building and working fine without modifications.

# Response on the Zephyr side

This module is consumed by https://github.com/xen-troops/meta-xt-prod-devel-rpi5
, which derives from Zephyr 3.6.0 and receives feature updates on the branch https://github.com/xen-troops/zephyr/tree/zephyr-v3.6.0-xt
On the Zephyr side, we will cherry-pick this change and continue the upstream integration toward main. https://github.com/zephyrproject-rtos/zephyr/pull/91643)


soburi:
@firscity @klogg @harunobukurokawa

Due to my mistake, I deleted my fork repository, which resulted in the #102 being closed, so I recreated it.

soburi:
@firscity 

https://github.com/zephyrproject-rtos/zephyr/pull/103152 will probably be included as soon as the 4.4.0 release is complete, so let's resume this discussion.

https://github.com/xen-troops/zephyr-xenlib/pull/103#issuecomment-3612073819

>    Next is regarding Xen public headers: we decide to proceed with your proposal to make zephyr-xenlib a module of zephyrproject-rtos repo. When it will be done in current implementation, we will move all public headers to zephyr-xenlib. I believe it will be done as a "initial" commit with some Xen Release public headers as is, we may discuss which one we should peek.

First, as a starting point for discussion, I've created a simple vanilla version with the files moved.

include/xen/public/domctl.h
```diff
@@ -0,0 +1,534 @@
...
+/* Should we expose the vPMU to the guest? */
+#define XEN_DOMCTL_CDF_vpmu		(1U << 7)
+
+/* Max XEN_DOMCTL_CDF_* constant.  Used for ABI checking. */
+#define XEN_DOMCTL_CDF_MAX		XEN_DOMCTL_CDF_vpmu
+
+	uint32_t flags;
+
+#define _XEN_DOMCTL_IOMMU_no_sharept	0
+#define XEN_DOMCTL_IOMMU_no_sharep	(1U << _XEN_DOMCTL_IOMMU_no_sharept)
```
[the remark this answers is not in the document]
soburi: For now, the policy is to leave things as they are.

include/xen/public/io/console.h
```diff
@@ -0,0 +1,48 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ * Copyright (c) 2005, Keir Fraser
+ */
+
+#ifndef __XEN_PUBLIC_IO_CONSOLE_H__
+#define __XEN_PUBLIC_IO_CONSOLE_H__
+
```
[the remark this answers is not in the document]
soburi: For now, the policy is to leave things as they are.

include/xen/public/hvm/ioreq.h
```diff
@@ -0,0 +1,125 @@
...
+ * 63....48|47..40|39..35|34..32|31........0
+ * SEGMENT |BUS   |DEV   |FN    |OFFSET
+ */
+struct ioreq {
+	uint64_t addr;     /* physical address */
+	uint64_t data;     /* data (or paddr of data) */
+	uint32_t count;    /* for rep prefixes */
+	uint32_t size;     /* size in bytes */
+	uint32_t vp_eport; /* evtchn for notifications to/from device model */
+	uint16_t _pad0;
```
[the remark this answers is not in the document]
soburi: For now, the policy is to leave things as they are.

include/xen/public/io/console.h
```diff
@@ -0,0 +1,48 @@
...
+	char out[2048];
+	XENCONS_RING_IDX in_cons, in_prod;
+	XENCONS_RING_IDX out_cons, out_prod;
+};
+
+#ifdef XEN_WANT_FLEX_CONSOLE_RING
+#include "ring.h"
+DEFINE_XEN_FLEX_RING(xencons);
+#endif
+
```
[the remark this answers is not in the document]
soburi: For now, the policy is to leave things as they are.

firscity:
Hi @soburi, thank you for your work and sorry for late reply!

Regarding public header migration - per my understanding, we have 2 final goals:
1. Make zephyr-xenlib a part of zephyrproject-rtos;
2. Move public header to zephyr-xenlib (and remove from Zephyr RTOS include/)

Also, it may be connected with 2 other issues:
- xen drivers are located inside Zephyr RTOS repo and I don't now if it is correct to use external headers from internal driver
- public headers is from Xen RELEASE-4.19.0 and it is good time to update it during transfer

Please, let me know your view on following points and how you see further steps.

CC: @drozd-svtln

soburi:
> Hi @soburi, thank you for your work and sorry for late reply!
> 
> Regarding public header migration - per my understanding, we have 2 final goals:
> 
>     1. Make zephyr-xenlib a part of zephyrproject-rtos;
> 
>     2. Move public header to zephyr-xenlib (and remove from Zephyr RTOS include/)
> 
> 
> Also, it may be connected with 2 other issues:
> 
>     * xen drivers are located inside Zephyr RTOS repo and I don't now if it is correct to use external headers from internal driver
> 
>     * public headers is from Xen RELEASE-4.19.0 and it is good time to update it during transfer
> 
> 
> Please, let me know your view on following points and how you see further steps.
> 
> CC: @drozd-svtln

>     * xen drivers are located inside Zephyr RTOS repo and I don't now if it is correct to use external headers from internal driver

I believe this is not a problem.
That is, hal is naturally used by the drivers, and picolibc is referenced by all modules.


>     * public headers is from Xen RELEASE-4.19.0 and it is good time to update it during transfer

I agree with this as well. However, I think the migration itself should be divided into two phases. That is, the first phase should be simple files moving and integrating modules (This PR), followed by the second phase of updating the headers.
It would be troublesome if changes were made together during module integration.

The introduction of the Xen module was previously proposed and approved. Re-issuing this proposal should minimize confusion. https://github.com/zephyrproject-rtos/zephyr/issues/91644

Fortunately or unfortunately, we won't be able to move forward until April 16th, when the 4.4.0 release is complete.
Let's think about it during that time.

firscity:
> I believe this is not a problem.
That is, hal is naturally used by the drivers, and picolibc is referenced by all modules.

Oh, I see, I'll check it and maybe try on my side. I will be happy to find a sequence of actions for seamless migration (like integrating zephyr-xenlib to main repo -> change includes to zephyr-xenlib -> remove public headers from Zephyr includes or smth).

> I agree with this as well. However, I think the migration itself should be divided into two phases. That is, the first phase should be simple files moving and integrating modules (This PR), followed by the second phase of updating the headers.
It would be troublesome if changes were made together during module integration.
The introduction of the Xen module was previously proposed and approved. Re-issuing this proposal should minimize confusion. https://github.com/zephyrproject-rtos/zephyr/issues/91644

Yeah, I remembers about TSC approval, but it was only regarding Xen headers, not about the whole zephyr-xenlib and, as a result, new repo on zephyrproject-rtos. Will it require any separate approval? Unfortunately, I'm a bit distant from Zephyr meetings and current practicies...

> Fortunately or unfortunately, we won't be able to move forward until April 16th, when the 4.4.0 release is complete.
Let's think about it during that time.

I believe fortunately, because due to high load on other work activities I can't spent much time on this, but hope I will finally have time to review and integrate Xenstore-client interface PR...

Thank you for your work and help!

soburi:
@firscity 

Sorry for the late response, I've been a bit busy.

> Oh, I see, I'll check it and maybe try on my side. I will be happy to find a sequence of actions for seamless migration (like integrating zephyr-xenlib to main repo -> change includes to zephyr-xenlib -> remove public headers from Zephyr includes or smth).

Regarding this, IMO,
1. Create a branch for migrating zephyr-xenlib.
2. Request to create a fork on the Zephyr side based on that branch. (Updating Issue https://github.com/zephyrproject-rtos/zephyr/issues/91644)
3. Update zephyr's west.yml to reference the created fork.

So, I would like you to create a branch from the current main for the purpose of 1. I will resubmit this PR for new one.

> Yeah, I remembers about TSC approval, but it was only regarding Xen headers, not about the whole zephyr-xenlib and, as a result, new repo on zephyrproject-rtos. Will it require any separate approval? Unfortunately, I'm a bit distant from Zephyr meetings and current practicies...

Of course, further explanation will be necessary, but reporting the result of scancode should suffice.

firscity:
Hi @soburi,

Sorry for delays in replays, I checked both this and #103 PRs and have some comments. I like the approach with Xen public headers in zephyr-xenlib, that will be a part of zephyrproject-rtos repository. It will allow us to use it in both zephyr-xenlib and Zephyr RTOS drivers. But I'm not sure if we should take all of the history for these headers in the way it is done in this PR.

In my opinion, we should create a separate directory in zephyr-xenlib, where we will place Xen public headers as is from certain Xen version (e.g. RELEASE-4.21.0). Then change small parts (like make INTERFACE_VERSION configurable from Zephyr Kconfigs etc) in a separate commit and switch all of the sources to use them from created directory as is. It will be easier to track and update them between Xen Releases, which will be necessary when we will became a part of zephyrproject-rtos I believe. 

Such approach will simplify this PR and also will not require to add new headers in #103.

I want to hear your and probably @lorc thoughts about this here and if everyone agree, this will unblock merging of this and Xenstore related PRs. After this we can start a process to move to another repo.

Thank you for your work and for waiting!

soburi:
> In my opinion, we should create a separate directory in zephyr-xenlib, where we will place Xen public headers as is from certain Xen version (e.g. RELEASE-4.21.0). Then change small parts (like make INTERFACE_VERSION configurable from Zephyr Kconfigs etc) in a separate commit and switch all of the sources to use them from created directory as is. It will be easier to track and update them between Xen Releases, which will be necessary when we will became a part of zephyrproject-rtos I believe.

I've rewritten it according to this policy. Could you take a look?

include/xen/public/hvm/hvm_op.h
```diff
@@ -142,15 +142,15 @@ typedef struct xen_hvm_xentrace xen_hvm_xentrace_t;
 DEFINE_XEN_GUEST_HANDLE(xen_hvm_xentrace_t);
 
 /* Following tools-only interfaces may change in future. */
-#if defined(__XEN__) || defined(__XEN_TOOLS__)
+#ifdef CONFIG_XEN_DOM0
```
firscity: Probably this is a good time to finally make smth like:
```
#ifdef CONFIG_XEN_DOM0
#define __XEN_TOOLS__
#endif
```
to minimize changes to headers. @lorc what do you think?

firscity (commented):
Only 1 comment, overall it looks great, thank you for your work!
Lets wait for @lorc before changing anything. After merging this one we will unblock all others, that are blocked by public headers.

include/xen/public/hvm/hvm_op.h
soburi: > Probably this is a good time to finally make smth like:
> 
> ```
> #ifdef CONFIG_XEN_DOM0
> #define __XEN_TOOLS__
> #endif
> ```
> 
> to minimize changes to headers. @lorc what do you think?

I'm completely agree it, and I updated it.
The headers keep as original without no modification.
It's quite cool.

CMakeLists.txt
```diff
@@ -3,6 +3,10 @@
 # Copyright (c) 2023 EPAM Systems
 # Author: Volodymyr Babchuk <volodymyr_babchuk@epam.com>
 
+set(ZEPHYR_XENLIB_DIR ${ZEPHYR_CURRENT_MODULE_DIR})
+
+zephyr_include_directories_ifdef(CONFIG_XEN ${ZEPHYR_XENLIB_DIR}/include)
```
firscity: Nit: I don't think that `..._ifdef()` is needed here, I think no one will try to build it without `CONFIG_XEN`, so may be unconditional `zephyr_include_directories()`

firscity (approved):
W or wo my comment fixed, please add my `Reviewed-by: Dmytro Firsov <dmytro_firsov@epam.com>` to all commits.

CMakeLists.txt
soburi: When enabled as a Zephyr module, this CMakeLists.txt file is always parsed.
Since there may be cases where builds are performed without Xen enabled in the same environment, it's helpful to leave this condition in place.

However, with Zephyr modules, it's common practice to enclose the entire code in if...endif statements, so I've modified it to follow that convention.

soburi:
> W or wo my comment fixed, please add my `Reviewed-by: Dmytro Firsov <dmytro_firsov@epam.com>` to all commits.

Fixed it.
I forgot about it. Thank you for pointing it out.

soburi:
@lorc 
If you have the time, I would appreciate it if you could check it.

include/xen/public/io/libxenvchan.h
xakep-amatop, line 1: Non-blocking note: the previous local `zephyr-xenlib` copy of `libxenvchan.h` had an include guard, while the newly imported shared copy matches upstream Xen and has no guard.

I do not think we should add a Zephyr-local delta here if the goal is to keep Xen public headers close to upstream. However, this is a small include-hygiene regression compared to the previous local copy, so it may be worth fixing in Xen upstream and then importing the fixed header later.

CMakeLists.txt
```diff
@@ -3,11 +3,19 @@
...
 
-add_subdirectory_ifdef(CONFIG_XEN_STORE_COMMON xenstore-common)
-add_subdirectory_ifdef(CONFIG_XEN_STORE_SRV xenstore-srv)
-add_subdirectory_ifdef(CONFIG_XEN_DOMAIN_MANAGEMENT xen-dom-mgmt)
-add_subdirectory_ifdef(CONFIG_XEN_SHELL xen-shell-cmd)
-add_subdirectory_ifdef(CONFIG_XEN_CONSOLE_SRV xen-console-srv)
-add_subdirectory_ifdef(CONFIG_XEN_VCH vch)
-add_subdirectory_ifdef(CONFIG_XEN_LIBFDT libfdt)
-add_subdirectory_ifdef(CONFIG_XSTAT xstat)
+if(CONFIG_XEN)
```
xakep-amatop, line 6: The top-level CMake now makes the whole module conditional on `CONFIG_XEN`.
That is probably fine if all xenlib features are intended to be usable only
when Zephyr's Xen support is enabled.

Should the Kconfig options mirror that by adding `depends on XEN` to the
module features, or by wrapping the xenlib menu in `if XEN`? Otherwise options
such as `XEN_STORE_SRV`, `XEN_DOMAIN_MANAGEMENT`, `XEN_VCH`, or `XSTAT` remain
selectable even though the module sources/include path are ignored when
`CONFIG_XEN=n`.

include/xen/public/io/libxenvchan.h
soburi: Thank you for your review.
It is undesirable there are differences with release version,
so I think it is good to correct in upstream, and incorporate at next update.

CMakeLists.txt
soburi: It make a sense. I updated it.

And `mainmenu` element not be here, removed it.

lorc (approved):
Thank you, Tokita-san,


xakep-amatop:

xakep-amatop (approved):

PR #106, opened 2025-11-20
xen-shell-cmd: Add options for shell for xenstore-srv and domain-mgmt.

soburi:
The shell function for xenstore server and domain management can now be switched using Kconfig
so that it can be disabled if not needed.
The default value is `y` for compatibility reasons.


firscity (approved):
Thank you for your work!

PR #117, opened 2026-07-22
[WIP] xenstore: client implementation + minor server bugfixes

svlad-90:
# Automated testing results:

[xenstore_client_api.html](https://github.com/user-attachments/files/30263490/xenstore_client_api.html)

# Zephyr XenStore Client PR Note

## Summary

This PR adds a Zephyr XenStore client and the missing dom0less XenStore setup
needed for Zephyr Dom0/DomU XenStore communication.

The series has three main parts:

1. Dom0less XenStore page setup in `xen-dom-mgmt`
2. A Zephyr DomU XenStore client
3. Focused XenStore server/common fixes needed by the client validation

The first commit handles the dom0less handoff that Linux normally performs via
`init-dom0less`: Dom0 must make sure the guest has a real XenStore shared page
and that the XenStore server and guest client agree on the same `STORE_PFN` and
`STORE_EVTCHN`.

This is needed for both Xen 4.19-style and newer Xen 4.22-style setups, but
for slightly different reasons. With Xen 4.19, `STORE_PFN` may still be
`~0ULL`, meaning Xen provided the event-channel doorbell but did not provide a
usable XenStore shared page yet; Zephyr Dom0 must allocate and publish that
page before the guest client can talk to the server. With newer Xen, Xen may
already publish a real `STORE_PFN`; Zephyr must reuse that exact page instead
of deriving another page from legacy magic-page assumptions. In both cases the
fix makes Zephyr follow the guest-visible `STORE_EVTCHN` / `STORE_PFN`
contract so the Dom0 server and DomU client use the same mailbox and doorbell.

The client then maps the XenStore ring, binds the event channel, performs the
`RECONNECT -> CONNECTED` handshake, and exposes a public C API for the XenStore
operations currently validated against the local Zephyr XenStore server.

The transport was later extended with a dedicated RX thread. This is needed
because XenStore replies and asynchronous watch events arrive on the same
shared response ring. The RX thread is the single reader of that ring: it
routes normal replies by `req_id` to the waiting caller and dispatches watch
events to the registered callback.

## Supported Client Operations

The public client API currently supports:

| XenStore operation | Client API | Status |
| --- | --- | --- |
| Connect / handshake | `xs_client_connect()` | Supported |
| Connection state query | `xs_client_is_connected()` | Supported |
| `XS_READ` | `xs_client_read()` | Supported |
| `XS_WRITE` | `xs_client_write()` | Supported |
| `XS_DIRECTORY` | `xs_client_directory()` | Supported |
| `XS_MKDIR` | `xs_client_mkdir()` | Supported |
| `XS_RM` | `xs_client_rm()` | Supported |
| `XS_GET_PERMS` | `xs_client_get_perms()` | Supported |
| `XS_SET_PERMS` | `xs_client_set_perms()` | Supported |
| `XS_WATCH` | `xs_client_watch()` | Supported |
| `XS_UNWATCH` | `xs_client_unwatch()` | Supported |
| `XS_RESET_WATCHES` | `xs_client_reset_watches()` | Supported |
| `XS_GET_DOMAIN_PATH` | `xs_client_get_domain_path()` | Supported |
| Watch event delivery | `xs_client_set_watch_callback()` | Supported, single process-wide callback |

## Intentionally Not Exposed

The client does not expose transaction helpers yet.

The local Zephyr XenStore server accepts transaction message types, but it does
not implement real transaction semantics such as staging, commit, abort, or
rollback. Exposing transaction-aware client APIs would therefore promise
behavior the validated server side does not provide.

The client also does not expose `XS_CONTROL`.

The Zephyr server currently treats `XS_CONTROL` as a shallow transport/probe
acknowledgement, not as a real control-command implementation. Keeping it out
of the public client API avoids making this look like supported command
semantics.

## Watch Callback Model

The client supports multiple XenStore watch registrations, because XenStore
itself identifies watch events by `path + token`.

The current Zephyr API exposes one process-wide callback:

```c
void xs_client_set_watch_callback(xs_client_watch_cb_t cb, void *user_data);
```

Applications that register multiple watches should dispatch by token inside
that callback.

Callback replacement is synchronized. If an application thread replaces or
clears the callback while the RX thread is already executing it, the setter
waits for the active callback to finish. If the callback replaces itself from
the RX thread, it updates immediately to avoid self-deadlock.

## Validation

Runtime validation was run in QEMU/Xen with the Zephyr XenStore server in Dom0
and the Zephyr XenStore client in DomU.

Result:

```text
44/44 PASS
XENSTORE CLIENT TEST PASS
```

The generated validation report is:

```text
zephyr-xenstore-client/dev/qemu-xen-zephyr-dom0-validation/reports/xenstore_client_api.html
```

The validation covers positive and negative cases for the public client API,
including connection, read/write, directory, mkdir, remove, permissions,
watches, watch callback delivery, reset watches, domain path lookup, and
concurrent callers.

----

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ if(CONFIG_XEN)
 
   add_subdirectory_ifdef(CONFIG_XEN_STORE_COMMON xenstore-common)
   add_subdirectory_ifdef(CONFIG_XEN_STORE_SRV xenstore-srv)
+  add_subdirectory_ifdef(CONFIG_XEN_STORE_CLIENT xenstore-client)
   add_subdirectory_ifdef(CONFIG_XEN_DOMAIN_MANAGEMENT xen-dom-mgmt)
   add_subdirectory_ifdef(CONFIG_XEN_SHELL xen-shell-cmd)
   add_subdirectory_ifdef(CONFIG_XEN_CONSOLE_SRV xen-console-srv)
--- a/Kconfig
+++ b/Kconfig
@@ -23,6 +23,26 @@ config XEN_STORE_SRV_SHELL_CMDS
 	help
 	  Enable shell for XenStore server.
 
+config XEN_STORE_CLIENT
+	bool "Enable XenStore client (frontend)"
+	select XEN_STORE_COMMON
+	help
+	  Enable the xenstore client/frontend library, for a DomU (guest) that
+	  needs to talk to the xenstore server. It maps the guest's own xenstore
+	  ring (HVM_PARAM_STORE_PFN), completes the connection handshake
+	  (RECONNECT -> CONNECTED, as required for dom0less guests whose ring Xen
+	  initialises to XENSTORE_RECONNECT), binds the store event channel
+	  (HVM_PARAM_STORE_EVTCHN) and provides basic read/write transactions.
+
+config XEN_STORE_CLIENT_AUTO_CONNECT
+	bool "Connect the XenStore client automatically at boot"
+	depends on XEN_STORE_CLIENT
+	default y
+	help
+	  Run xs_client_connect() from a SYS_INIT hook at boot. The hook is a
+	  no-op on domains without a provisioned store event channel (e.g. the
+	  xenstore domain itself), so it only acts on real client guests.
+
 config XEN_LIBFDT
 	bool "Enable libfdt support"
 	help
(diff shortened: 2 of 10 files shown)


svlad-90:
Do not review yet. I need to consider changes proposed in:
https://github.com/xen-troops/zephyr-xenlib/pull/103
https://github.com/xen-troops/zephyr-xenlib/pull/110

svlad-90:
I've reviewed https://github.com/xen-troops/zephyr-xenlib/pull/103 and left my findings. Let's try to make it the target implementation. I'll close this PR.

PR #119, opened 2026-07-23
ci: make fork PR builder checks reliable

svlad-90:
ci: use pull_request for PR builders

Run PR builders in the pull_request context so fork pull
request code is checked out with the restricted pull request token.

The previous pull_request_target workflow runs in the trusted base
repository context. That context refuses the unsafe fork-head checkout
pattern used by this workflow, which prevents fork PR validation from
starting.

----

ci: refresh apt dependencies for builders

Refresh the apt package index before installing runner packages and
pass -y so the install step can run in non-interactive CI.

Install cmake with the other host tools because Zephyr SDK setup
requires it before the project build starts.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-  pull_request_target:
+  pull_request:
 
 env:
   sdk_version: ${{ vars.SDK_VERSION  || '0.16.5-1' }}
@@ -79,7 +79,6 @@ jobs:
       - name: Check out zephyr-xenlib
         uses: actions/checkout@v4
         with:
-          ref: ${{ github.event.pull_request.head.sha }}
           path: zephyr-xenlib
 
       - name: Build zephyr-dom0-xt
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -32,7 +32,8 @@ jobs:
       - name: Install dependencies
         run: |
           pip install west pyelftools protobuf grpcio-tools
-          sudo apt-get install ninja-build protobuf-compiler gperf
+          sudo apt-get update
+          sudo apt-get install -y cmake ninja-build protobuf-compiler gperf
 
       - name: Cache Zephyr SDK
         id: cache-sdk


svlad-90:
Fix validated here: https://github.com/svlad-90/zephyr-xenlib/pull/3. The fix can't be validated on the original repo until it is merged.

image

3 issues fixed:

  1. Switched PR builds from pull_request_target to pull_request.
     This avoids running fork PR code in the trusted base-repository context and fixes GitHub’s unsafe checkout refusal.

  2. Installed missing cmake.
     Zephyr SDK setup/build requires it.

  3. Hardened Zephyr SDK download.
     Added retry logic and archive validation with tar -tf before extraction, so partial or corrupted downloads do not break the workflow.

  4. Refreshed apt metadata before installing packages.
     Added sudo apt-get update before apt-get install. This is mainly for local reproduction of the workflow with the act tool.

  5. Made apt installation non-interactive.
     Added -y to sudo apt-get install. This is mainly for local reproduction of the workflow with the act tool.

lorc (commented):
1. Please make atomic commits. By reading the commit message I can already see that there should be at least 3 different patches
2. Please refrain from unneeded whitespace changes

.github/workflows/build.yaml
```diff
@@ -47,9 +47,21 @@ jobs:
         run: |
           mkdir sdk
           cd sdk
-          wget -q https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v${{ env.sdk_version }}/zephyr-sdk-${{ env.sdk_version }}_linux-x86_64.tar.xz
-          tar xf zephyr-sdk-${{ env.sdk_version }}_linux-x86_64.tar.xz
-          rm zephyr-sdk-${{ env.sdk_version }}_linux-x86_64.tar.xz
+          sdk_archive=zephyr-sdk-${{ env.sdk_version }}_linux-x86_64.tar.xz
+          sdk_url=https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v${{ env.sdk_version }}/${sdk_archive}
+          for attempt in 1 2 3; do
```
lorc: Is this better than just `curl --retry=3` ?

.github/workflows/build.yaml
```diff
@@ -47,9 +47,21 @@ jobs:
...
+            rm -f "${sdk_archive}"
+            wget --progress=dot:giga "${sdk_url}" -O "${sdk_archive}"
+            if tar -tf "${sdk_archive}" >/dev/null; then
```
lorc: why are you using UNIX-style `tar` argument here...

.github/workflows/build.yaml
```diff
@@ -47,9 +47,21 @@ jobs:
...
+              break
+            fi
+            if [ "${attempt}" = 3 ]; then
+              echo "Downloaded Zephyr SDK archive is invalid after ${attempt} attempts"
+              exit 1
+            fi
+          done
+          tar xf "${sdk_archive}"
```
lorc: ... but traditional style here?

.github/workflows/build.yaml
svlad-90: Hi @lorc, removed this change. The SDK is already cached, so the download step only runs on cache misses, SDK version changes, or cache eviction. If that rare download step fails, rerunning the workflow should be enough.

.github/workflows/build.yaml
svlad-90: Hi @lorc, removed this change. The SDK is already cached, so the download step only runs on cache misses, SDK version changes, or cache eviction. If that rare download step fails, rerunning the workflow should be enough.

.github/workflows/build.yaml
svlad-90: Hi @lorc, removed this change. The SDK is already cached, so the download step only runs on cache misses, SDK version changes, or cache eviction. If that rare download step fails, rerunning the workflow should be enough.

lorc (approved):

PR #120, opened 2026-07-25
xen-dom-mgmt: Add include `zephyr/arch/cache.h`

soburi:
Add include `zephyr/arch/cache.h` to resolve `arch_dcache_flush_and_invd_range()`.


xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -7,6 +7,7 @@
 #undef _POSIX_C_SOURCE
 #define _POSIX_C_SOURCE 200809L
 
+#include <zephyr/arch/cache.h>
```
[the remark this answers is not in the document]
soburi: Fixed.

xen-dom-mgmt/src/xen-dom-mgmt.c
```diff
@@ -7,6 +7,7 @@
...
 #include <zephyr/cache.h>
```
firscity, line 11: @soburi is adding this one is not enough? @sa-kib please take a look, did it work for you during PR submission?

sa-kib: I saw that other users of this API just used `#include <zephyr/cache.h>` so I did the same. Worked for me.

If it doesn't work anymore perhaps we can use `sys_cache_data_flush_and_invd_range()` API? It is more generic and wrapped around in build config checks etc. I guess I should've used this API from the start..

firscity: @sa-kib, oh, I see. Let's wait for clarification from @soburi, I believe generic API is better that use arch-specific here, but let's hear from him

soburi: I agree it would be preferable to use a system call ultimately.
However, since we are currently in the middle of a version update, I believe it is best to avoid doing two things at once. IMO, it may better to proceed with the current approach for now and handle that update separately once the current one is complete.

firscity (approved):

firscity:
Or maybe we should move to https://github.com/zephyrproject-rtos/zephyr-xenlib?

soburi:
> Or maybe we should move to https://github.com/zephyrproject-rtos/zephyr-xenlib?

One major issue is that you cannot obtain permissions for the repository unless you register it in Zephyr's `west.yml`...
First, merge here, and once we have the permissions, update the main repository there with them.

soburi:
> `Acked-by: Dmytro Firsov <dmytro_firsov@epam.com>`

Applied. Thank you.

PR #123, opened 2026-08-01
include: Update Xen public header to RELEASE-4.22.0

soburi:
Update Xen public header to RELEASE-4.22.0.
Excluding xen-compat.h from it to keep modification for zephyr.


klogg:
@drozd-svtln please collaborate with @soburi on this

drozd-svtln:
Hi @soburi, changes in headers LGTM, but i found that we missed two files `arch-ppc.h` and `arch-riscv.h`. Is this intentional?

firscity:
@drozd-svtln IIRC they are under non-permissive license, thus were removed

drozd-svtln:
Ah, yes, I see. But those files are included in other headers, do we need to remove these includes or guard them somehow? Not in the scope of this PR, of course

soburi:
Yes. I make sure not to include files with incompatible licenses here. It seems we don't need to worry about PPC for the time being, but I would certainly appreciate it RISCV part if the license could be changed, if possible.

drozd-svtln:
```Reviewed-by: Svitlana Drozd <svitlana_drozd@epam.com>```

drozd-svtln (approved):

